PR #26, opened 2016-06-23
Adding 'flist' option - allows providing the input files in a text file

shachar-i:
Hi,

This commit is necessary for situations where one would like to assemble a volume texture from a very large number of images. It becomes cumbersome to list all the files in the command line and you can also very easily exceed the cmd command length limit of 8191 chars (that happened to me).

I added the common solution of using a text file which includes a list of file paths. This is compatible with all other usage scenarios.

Usage:
-flist <filename>   use text file with a list of input files (one per line)

Thanks for sharing this project, it's great!
Shachar

### DIFF
--- a/Texassemble/texassemble.cpp
+++ b/Texassemble/texassemble.cpp
@@ -37,6 +37,7 @@ enum OPTIONS    // Note: dwOptions below assumes 32 or less options.
     OPT_USE_DX10,
     OPT_NOLOGO,
     OPT_SEPALPHA,
+	OPT_FILELIST,
     OPT_MAX
 };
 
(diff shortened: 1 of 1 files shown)


walbourn:
As you've not signed the CLA, I'll close this PR but I'll look at implementing the ``flist`` option in my command-line tools based on your suggested implementation.  Thanks!

walbourn:
Reimplemented and [committed](https://github.com/Microsoft/DirectXTex/commit/3a4748ddeb143f2592536ca39aba6ed5e68eb5d6). Also updated DirectxTK, DirectXMesh, and UVAtlas tools.

PR #42, opened 2016-12-15
[TexConv] Add alpha coverage to mipmap generation.

CM-SamOates:
Alpha coverage increases alpha accuracy at lower mip levels by scaling the alpha channel based upon the alpha coverage of the original image.

Alpha coverage takes a ratio between 0 and 1, Any number greater than 1 disables alpha coverage.
The default value (if the command line option is not supplied) is 1, disabling the feature by default.

This technique allows noisy textures (such as leaves on a tree) to retain a high accuracy in the mips when compared against the original source image.

An overview of the technique implemented can be found here: http://the-witness.net/news/2010/09/computing-alpha-mipmaps/
However, some optimizations have been made, as well as using multi-sampling from the source image to give a smoother distribution within the alpha of the mips.

### DIFF
--- a/DirectXTex/DirectXTex.h
+++ b/DirectXTex/DirectXTex.h
@@ -274,6 +274,7 @@ namespace DirectX
 
         const TexMetadata& __cdecl GetMetadata() const { return m_metadata; }
         const Image* __cdecl GetImage(_In_ size_t mip, _In_ size_t item, _In_ size_t slice) const;
+        Image* __cdecl GetWritableImage(_In_ size_t mip, _In_ size_t item, _In_ size_t slice) const;
 
         const Image* __cdecl GetImages() const { return m_image; }
         size_t __cdecl GetImageCount() const { return m_nimages; }
@@ -471,7 +472,7 @@ namespace DirectX
     HRESULT __cdecl GenerateMipMaps( _In_ const Image& baseImage, _In_ DWORD filter, _In_ size_t levels,
                                      _Inout_ ScratchImage& mipChain, _In_ bool allow1D = false );
     HRESULT __cdecl GenerateMipMaps( _In_reads_(nimages) const Image* srcImages, _In_ size_t nimages, _In_ const TexMetadata& metadata,
-                                     _In_ DWORD filter, _In_ size_t levels, _Inout_ ScratchImage& mipChain );
+                                     _In_ DWORD filter, _In_ size_t levels, _Inout_ ScratchImage& mipChain, const float alphaCoverageRatio = 1.0f);
         // levels of '0' indicates a full mipchain, otherwise is generates that number of total levels (including the source base image)
         // Defaults to Fant filtering which is equivalent to a box filter
 
--- a/DirectXTex/DirectXTexMipmaps.cpp
+++ b/DirectXTex/DirectXTexMipmaps.cpp
@@ -2521,6 +2521,129 @@ namespace
 // Entry-points
 //=====================================================================================
 
+float clamp(float value, float min, float max)
+{
+    return value < min ? min : value > max ? max : value;
+}
+
+//-------------------------------------------------------------------------------------
+// Alpha to coverage, for mip generation
+//-------------------------------------------------------------------------------------
+_Use_decl_annotations_
+float CalculateAlphaCoverage(const Image& image, const float alphaCoverageRatio, const float scale)
+{
+    float coverage = 0.0f;
+
+    ScopedAlignedArrayXMVECTOR scanline(reinterpret_cast<XMVECTOR*>(_aligned_malloc((sizeof(XMVECTOR) * image.width), 16)));
+    const uint8_t *pPixels = image.pixels;
+    assert(pPixels);
+
+    const float width = static_cast<float>(image.width);
+    const float height = static_cast<float>(image.height);
+
+    const unsigned int n = 8;
+
+    for (size_t h = 0; h < image.height - 1; ++h)
+    {
+        if (!_LoadScanline(scanline.get(), image.width, pPixels, image.rowPitch, image.format))
+            continue;
+
+        XMVECTOR* currentRowPixels = scanline.get();
+
+        if (!_LoadScanline(scanline.get(), image.width, pPixels  + image.rowPitch, image.rowPitch, image.format))
+            continue;
+
+        XMVECTOR* nextRowPixels = scanline.get();
+
+        for (size_t w = 0; w < image.width - 1; ++w)
+        {            
+            // TODO: Use std::clamp when c++17 is available.
+            float alpha00 = clamp(currentRowPixels[w].m128_f32[3] * scale, 0.0f, 1.0f);
+            float alpha10 = clamp(currentRowPixels[w + 1].m128_f32[3] * scale, 0.0f, 1.0f);
+            float alpha01 = clamp(nextRowPixels[w].m128_f32[3] * scale, 0.0f, 1.0f);
+            float alpha11 = clamp(nextRowPixels[w + 1].m128_f32[3] * scale, 0.0f, 1.0f);
+
+            for (float fy = 0.5f / n; fy < 1.0f; fy++) {
+                for (float fx = 0.5f / n; fx < 1.0f; fx++) {
+                    float alpha = alpha00 * (1 - fx) * (1 - fy) + alpha10 * fx * (1 - fy) + alpha01 * (1 - fx) * fy + alpha11 * fx * fy;
+                    if (alpha > alphaCoverageRatio)
+                    {
+                        coverage += 1.0f;
+                    }
+                }
+            }
+        }
+
+        pPixels += image.rowPitch;
+    }
+
+    return coverage / float(width * height * n * n);
+}
+
+_Use_decl_annotations_
+void ScaleAlphaToCoverage(Image& image, const float alphaCoverageRatio, const float desiredCoverage)
+{
+    float minScale = 0.0f;
+    float maxScale = 4.0f;
+    float scale = 1.0f;
+
+    static const int s_iterationCount = 10;
+    for (int scalingIndex = 0; scalingIndex < s_iterationCount; ++scalingIndex)
+    {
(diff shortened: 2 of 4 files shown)


DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
 // Entry-points
 //=====================================================================================
 
+float clamp(float value, float min, float max)
```
walbourn: These helpers should be higher up (not the entry-points section) and in an anonymous namespace

walbourn: These new helper functions should be higher up (not the entry-points section) and in an anonymous namespace

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+{
+    return value < min ? min : value > max ? max : value;
+}
+
+//-------------------------------------------------------------------------------------
+// Alpha to coverage, for mip generation
+//-------------------------------------------------------------------------------------
+_Use_decl_annotations_
```
walbourn: This only works if you have a SAL annotated prototype. Since it's a locally declared helper, you need to SAL annotate it here.

You can validate this by using ``/analyze``.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2897,12 +3032,27 @@ HRESULT DirectX::GenerateMipMaps(
...
-            return hr;
+            break;
 
         default:
             return HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED);
         }
     }
+
+    // Scale alpha on all mips
+    if (hr == S_OK && alphaCoverageRatio < 1.0f)
```
walbourn: This only implements the 1D & 2D case. Is there no equivalent for 3D?

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+
+//-------------------------------------------------------------------------------------
+// Alpha to coverage, for mip generation
+//-------------------------------------------------------------------------------------
+_Use_decl_annotations_
+float CalculateAlphaCoverage(const Image& image, const float alphaCoverageRatio, const float scale)
+{
+    float coverage = 0.0f;
+
+    ScopedAlignedArrayXMVECTOR scanline(reinterpret_cast<XMVECTOR*>(_aligned_malloc((sizeof(XMVECTOR) * image.width), 16)));
```
walbourn: ``_aligned_malloc`` doesn't throw on failure, so you need to explicitly check for out-of-memory. If nothing else, it should silently fail right here to avoid a crash in out-of-memory scenarios.

    if (!scanline)
        return 0.f

A better approach would be to have the caller pre-allocated the scratch memory so you return an appropriate ``HRESULT`` failure on out-of-memory. You should also allocate the largest possible width and reuse it instead of current implementation which is allocating/freeing the memory over and over again.

DirectXTex/DirectXTexMipmaps.cpp
walbourn: Make ``clamp`` inline.

Texconv/texconv.cpp
```diff
@@ -1279,6 +1286,24 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
             case OPT_X2_BIAS:
                 dwConvert |= TEX_FILTER_FLOAT_X2BIAS;
                 break;
+
+            case OPT_ALPHA_COVERAGE:
+                if (swscanf_s(pValue, L"%f", &alphaCoverageRatio) != 1)
+                {
+                    wprintf(L"Invalid value specified with -alphacoverage (%s)\n", pValue);
+                    wprintf(L"\n");
```
walbourn: Just put the extra ``\n`` in the previous ``wprintf``. Same with the next one.

DirectXTex/DirectXTex.h
```diff
@@ -471,7 +472,7 @@ namespace DirectX
     HRESULT __cdecl GenerateMipMaps( _In_ const Image& baseImage, _In_ DWORD filter, _In_ size_t levels,
                                      _Inout_ ScratchImage& mipChain, _In_ bool allow1D = false );
     HRESULT __cdecl GenerateMipMaps( _In_reads_(nimages) const Image* srcImages, _In_ size_t nimages, _In_ const TexMetadata& metadata,
-                                     _In_ DWORD filter, _In_ size_t levels, _Inout_ ScratchImage& mipChain );
+                                     _In_ DWORD filter, _In_ size_t levels, _Inout_ ScratchImage& mipChain, const float alphaCoverageRatio = 1.0f);
```
walbourn: To avoid clutter I've not been marking simple parameter passed by value as ``const``.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2742,7 +2866,18 @@ HRESULT DirectX::GenerateMipMaps(
 
     assert(baseImages.size() == metadata.arraySize);
 
-    HRESULT hr;
+    // Determine the alpha coverage for each map.
+    std::vector<float> alphaCoverages;
+    if (alphaCoverageRatio < 1.0f)
+    {
+        alphaCoverages.reserve(baseImages.size());
+        for (auto imageIt = baseImages.begin(); imageIt != baseImages.end(); ++imageIt)
```
walbourn: ``cbegin``/``cend`` here?

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+    for (size_t h = 0; h < image.height; ++h)
+    {
+        if (!_LoadScanline(scanline.get(), image.width, pPixels, image.rowPitch, image.format))
+            continue;
+
+        XMVECTOR* currentRowPixels = scanline.get();
+        for (size_t w = 0; w < image.width; ++w)
+        {
+            XMVECTOR& rgba = currentRowPixels[w];
+            rgba.m128_f32[3] = std::min<float>(rgba.m128_f32[3] * scale, 1.0f);
```
walbourn: This won't build on ARM/ARM64 platforms. Use ``XMVectorGetW`` to access the element.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+    if (minVarience < curVarience && minVarience < maxVarience)
+    {
+        scale = minScale;
+    }
+    else if (maxVarience < minVarience && maxVarience < curVarience)
+    {
+        scale = maxScale;
+    }
+
+    ScopedAlignedArrayXMVECTOR scanline(reinterpret_cast<XMVECTOR*>(_aligned_malloc((sizeof(XMVECTOR) * image.width), 16)));
```
walbourn: As I noted above, I'd pass the allocated memory buffer to ``CalculateAlphaCoverage`` and then reuse it here.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+
+        if (!_LoadScanline(scanline.get(), image.width, pPixels  + image.rowPitch, image.rowPitch, image.format))
+            continue;
+
+        XMVECTOR* nextRowPixels = scanline.get();
+
+        for (size_t w = 0; w < image.width - 1; ++w)
+        {            
+            // TODO: Use std::clamp when c++17 is available.
+            float alpha00 = clamp(currentRowPixels[w].m128_f32[3] * scale, 0.0f, 1.0f);
```
walbourn: Again, won't build on ARM/ARM64. Use ``XMVectorGetW``.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+            float alpha10 = clamp(currentRowPixels[w + 1].m128_f32[3] * scale, 0.0f, 1.0f);
+            float alpha01 = clamp(nextRowPixels[w].m128_f32[3] * scale, 0.0f, 1.0f);
+            float alpha11 = clamp(nextRowPixels[w + 1].m128_f32[3] * scale, 0.0f, 1.0f);
+
+            for (float fy = 0.5f / n; fy < 1.0f; fy++) {
```
walbourn: Style-wise, the ``{`` should always start on a new line.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+    const float height = static_cast<float>(image.height);
+
+    const unsigned int n = 8;
+
+    for (size_t h = 0; h < image.height - 1; ++h)
+    {
+        if (!_LoadScanline(scanline.get(), image.width, pPixels, image.rowPitch, image.format))
+            continue;
+
+        XMVECTOR* currentRowPixels = scanline.get();
```
walbourn: const

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -2521,6 +2521,129 @@ namespace
...
+
+        if (!_LoadScanline(scanline.get(), image.width, pPixels  + image.rowPitch, image.rowPitch, image.format))
+            continue;
+
+        XMVECTOR* nextRowPixels = scanline.get();
```
walbourn: const

walbourn:
At this point I'm just blocked on you signing the CLA to accept this pull request...

CM-SamOates:
I've been waiting on my companies lawyers for almost a year, I can't see them ever getting back to me.
As such I'll have to close this request.

walbourn:
Ah, sorry to hear that. If things change, feel free to resync and then resubmit a new PR.

> Note you could always try a bit of shaming by telling them that they are now less friendly to open source than Microsoft :)

PR #48, opened 2017-03-28
Added support for linear color space PNG writing

noirsoft:
I noticed that if srgb is not set for the texture formet, the PNG writer will still put an sRGB/RenderingIntent into the PNG header, and will not have a correct gamma value of 1.0. I added code that works for my application.

### DIFF
--- a/DirectXTex/DirectXTexWIC.cpp
+++ b/DirectXTex/DirectXTexWIC.cpp
@@ -588,6 +588,17 @@ namespace
                     value.bVal = 0;
                     (void)metawriter->SetMetadataByName(L"/sRGB/RenderingIntent", &value);
                 }
+                // linear space.
+                else
+                {
+                    // add gAMA chunk with gamma 1.0
+                    value.vt = VT_UI4;
+                    value.uintVal = 100000; // gama value * 100,000 -- i.e. gamma 1.0
+                    (void)metawriter->SetMetadataByName( L"/gAMA/ImageGamma", &value );
+
+                    // remove sRGB chunk which is added by default.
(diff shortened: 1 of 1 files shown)


noirsoft:
We haven't needed to test other image type code paths. We are using DirectXTex to test our rendering pipeline by dumping out various intermediate render stages and comparing to reference images as part of an automated test suite.  I believe that most viewers, in absence of an explicit gAMA chunk, will assume sRGB gamma. (Some, like IE and the windows photo viewer, always assume sRGB gamma, even with the gAMA chunk)
I may have some spare cycles this week to test, so to make sure I'm testing what you want, I'll list what I think I should test this week, please let me know if there are other test cases you want explored.
1) Test removing the sRGB chunk but no gAMA chunk in PNG
2) see what existing functionality does with jpeg and tiff output with a 16 bit linear texture input.
Is that right?

noirsoft:
I will try to run some tests in the next few days and will get back to you with the results.
Ideally we would be able to specify the gamma value, but for our usage, we know it's supposed to be gamma 1.0, and our reference images do have a gAMA 1.0 chunk set.

DirectXTex/DirectXTexWIC.cpp
```diff
@@ -588,6 +588,17 @@ namespace
                     value.bVal = 0;
                     (void)metawriter->SetMetadataByName(L"/sRGB/RenderingIntent", &value);
                 }
+                // linear space.
```
walbourn: If this is the proper behavior, I should also update the ``ScreenGrab`` modules in _DirectX Tool Kit_ for DX11 and DX12.

PR #65, opened 2017-11-15
Fix compiler errors (Clang 5.0.0)

darwikey:
Hello,
I'm trying to compile DirectXTex with Clang 5 (inside Visual Studio 2017) but I'm getting errors:

In file included from DirectXTex\DirectXTex_Z.cpp:15:
DirectXTex/DirectXTexConvert.cpp(3851,9): error : use of overloaded operator '-' is ambiguous (operand type 'const DirectX::XMVECTORF32')

I propose this fix, but maybe a better thing to do is to fix FXMVECTOR operators ?

### DIFF
--- a/DirectXTex/Filters.h
+++ b/DirectXTex/Filters.h
@@ -201,7 +201,7 @@ inline void _CreateCubicFilter(_In_ size_t source, _In_ size_t dest, _In_ bool w
     a1 = XMVectorSubtract( a1, XMVectorMultiply( g_cubicSixth, d3 ) ); \
     XMVECTOR a2 = XMVectorAdd( XMVectorMultiply( g_cubicHalf, d0 ), XMVectorMultiply( g_cubicHalf, d2 ) ); \
     XMVECTOR a3 = XMVectorSubtract( XMVectorMultiply( g_cubicSixth, d3 ), XMVectorMultiply( g_cubicSixth, d0 ) ); \
-    d3 = XMVectorSubtract( d3, XMVectorMultiply( g_cubicHalf, d2 ) ); \
+    a3 = XMVectorSubtract( a3, XMVectorMultiply( g_cubicHalf, d2 ) ); \
     XMVECTOR vdx = XMVectorReplicate( dx ); \
     XMVECTOR vdx2 = XMVectorMultiply( vdx, vdx ); \
     XMVECTOR vdx3 = XMVectorMultiply( vdx2, vdx ); \
--- a/DirectXTex/Filters.h
+++ b/DirectXTex/Filters.h
@@ -102,15 +102,17 @@ inline void _CreateLinearFilter(_In_ size_t source, _In_ size_t dest, _In_ bool 
 }
 
 #define BILINEAR_INTERPOLATE( res, x, y, r0, r1 ) \
-    res = ( y.weight0 * ( (r0)[ x.u0 ] * x.weight0 + (r0)[ x.u1 ] * x.weight1 ) ) \
-          + ( y.weight1 * ( (r1)[ x.u0 ] * x.weight0 + (r1)[ x.u1 ] * x.weight1 ) )
+    res = XMVectorAdd( XMVectorScale( XMVectorAdd( XMVectorScale( (r0)[ x.u0 ], x.weight0 ), XMVectorScale( (r0)[ x.u1 ], x.weight1 ) ), y.weight0 ), \
+                       XMVectorScale( XMVectorAdd( XMVectorScale( (r1)[ x.u0 ], x.weight0 ), XMVectorScale( (r1)[ x.u1 ], x.weight1 ) ), y.weight1 ) )
 
 #define TRILINEAR_INTERPOLATE( res, x, y, z, r0, r1, r2, r3 ) \
-    res = ( z.weight0 * ( ( y.weight0 * ( (r0)[ x.u0 ] * x.weight0 + (r0)[ x.u1 ] * x.weight1 ) ) \
-                          + ( y.weight1 * ( (r1)[ x.u0 ] * x.weight0 + (r1)[ x.u1 ] * x.weight1 ) ) ) ) \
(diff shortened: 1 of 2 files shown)


walbourn:
Last year I updated DirectXMath to compile with clang. One of the issues I encountered is that clang does not define __m128 as a struct the way Visual C++ and the Intel compiler do. Instead it's defined as an opaque type. Because of C++ rules you're not able to overload just on the type. Therefore I added logic to remove those overloads when building with clang. Obviously that means any client code that relies on those overloads has to be updated as you've done.

I'll review the changes you proposed and make sure all the code still builds  run my tests, and see if I can get it into the main line.

DirectXTex/Filters.h
```diff
@@ -195,9 +195,9 @@ inline void _CreateCubicFilter(_In_ size_t source, _In_ size_t dest, _In_ bool w
     XMVECTOR d0 = (p0) - a0; \
     XMVECTOR d2 = (p2) - a0; \
     XMVECTOR d3 = (p3) - a0; \
-    XMVECTOR a1 = d2 - g_cubicThird*d0 - g_cubicSixth*d3; \
-    XMVECTOR a2 = g_cubicHalf*d0 + g_cubicHalf*d2; \
-    XMVECTOR a3 = g_cubicSixth*d3 - g_cubicSixth*d0 - g_cubicHalf*d2;  \
+    XMVECTOR a1 = d2 - XMVectorMultiply( g_cubicThird, d0 ) - XMVectorMultiply( g_cubicSixth, d3 ); \
```
walbourn: Why didn't you have to use XMVectorSubtract( here? How does operator-(FXVMECTOR, FXMVECTOR) work here?

walbourn:
I tried building the code with clang/C2 using the latest Windows 10 SDK. I get the following:

   1>In file included from C:\Program Files (x86)\Windows Kits\10\Include\10.0.16299.0\um/d3d11_1.h:127:
    1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.16299.0\um/d3d11.h(1064,48): error : default initialization of an object of const type 'const CD3D11_DEFAULT' without a user-provided default constructor

How did you work around that? Did you use ``D3D11_NO_HELPERS``?

DirectXTex/Filters.h
walbourn: All the code around diffuse-errors also needs ``operator +=(FMXVECTOR)`` which also fails to build.

walbourn: In ``filters.h``, the macros ``BILINEAR_INTERPOLATE`` and ``TRILINEAR_INTERPOLATE`` also fail to build without the overloads.

darwikey: It's strange I succeed to compile, what toolset are you using ?

walbourn:
Looks good generally. I noticed that in DirectXTexWIC.cpp with clang/c2, you need to change:

            VARIANT varValue;
            varValue.vt = VT_BOOL;
            varValue.boolVal = VARIANT_TRUE;
            (void)props->Write(1, &option, &varValue);

to

            VARIANT varValue;
            varValue.n1.n2.vt = VT_BOOL;
            varValue.n1.n2.n3.boolVal = VARIANT_TRUE;
            (void)props->Write(1, &option, &varValue);

For clang to be able to build this code...

darwikey:
No I didn't encounter that issue

walbourn:
Something is off in the ``CUBIC_INTERPOLATE``. With the new code a bunch of my cubic filtering tests are failing...

DirectXTex/Filters.h
```diff
@@ -192,16 +194,18 @@ inline void _CreateCubicFilter(_In_ size_t source, _In_ size_t dest, _In_ bool w
...
-    XMVECTOR a2 = g_cubicHalf*d0 + g_cubicHalf*d2; \
-    XMVECTOR a3 = g_cubicSixth*d3 - g_cubicSixth*d0 - g_cubicHalf*d2;  \
+    XMVECTOR d0 = XMVectorSubtract( p0, a0 ); \
+    XMVECTOR d2 = XMVectorSubtract( p2, a0 ); \
+    XMVECTOR d3 = XMVectorSubtract( p3, a0 ); \
+    XMVECTOR a1 = XMVectorSubtract( d2, XMVectorMultiply( g_cubicThird, d0 ) ); \
+    a1 = XMVectorSubtract( a1, XMVectorMultiply( g_cubicSixth, d3 ) ); \
+    XMVECTOR a2 = XMVectorAdd( XMVectorMultiply( g_cubicHalf, d0 ), XMVectorMultiply( g_cubicHalf, d2 ) ); \
+    XMVECTOR a3 = XMVectorSubtract( XMVectorMultiply( g_cubicSixth, d3 ), XMVectorMultiply( g_cubicSixth, d0 ) ); \
+    d3 = XMVectorSubtract( d3, XMVectorMultiply( g_cubicHalf, d2 ) ); \
```
walbourn: d3 here should be a3 in both places!

walbourn:
Found it:

    XMVECTOR a3 = XMVectorSubtract( XMVectorMultiply( g_cubicSixth, d3 ), XMVectorMultiply( g_cubicSixth, d0 ) ); \ 
    d3 = XMVectorSubtract( d3, XMVectorMultiply( g_cubicHalf, d2 ) ); \ 

->

    XMVECTOR a3 = XMVectorSubtract( XMVectorMultiply( g_cubicSixth, d3 ), XMVectorMultiply( g_cubicSixth, d0 ) ); \ 
    a3 = XMVectorSubtract( a3, XMVectorMultiply( g_cubicHalf, d2 ) ); \ 

The ``a3`` term was accidentally made into ``d3`` in the translation.

Fix that typo and I should be able to merge your PR.

darwikey:
I fixed it, I'm really sorry, I have double checked the code but I missed that

walbourn:
No problem. It's an error-prone process which is why I hadn't done it yet...  Thanks!

PR #81, opened 2018-03-05
Reduce quantize drifting

elasota:
Improve accuracy of endpoint quantization.  Old version has upward drift that gets more severe at higher values.

If you need an example of a type of image that causes this problem, just fill an image with the color (243, 243, 243, 255) and encode it with -bcmax.  The decoded output will be filled with (245, 243, 245, 255).  With the fix, the decoded output produces the same color.


walbourn:
I'll review this as well...

walbourn:
I'll take this PR once I have my BC validator test done... I've been focused on VS 2017 (15.7 update) and Windows 10 SDK (17134) the past month, so I'm getting back to this...

walbourn:
Fixed in [this commit](https://github.com/Microsoft/DirectXTex/commit/707d9acc4973f8458164614f9dd1275aadac21ae).

Thanks!

PR #82, opened 2018-03-22
[texconv] Implemented alpha coverage preservation option

jmuulian:
(Note: I noticed after first implementing these changes that @CM-SamOates had opened a PR for this feature, but it looks like that has been closed now so I'll go ahead with this PR.)

These changes implement mipmap alpha coverage preservation, which is a useful technique to solve a common rendering problem where lower resolution mipmaps cause alpha tested/blended objects to shrink/disappear with distance.

@castano explains the problem and solution technique here: http://the-witness.net/news/2010/09/computing-alpha-mipmaps/

My implementation is adapted from the NVTT code referenced in the article above (which is in the public domain), with some minor fixes and some texconv-specific optimizations.

I've tried to follow conventions established elsewhere in the codebase, but note that I've only built and tested on Win32/x64.

I expect the feature to work correctly with all existing features, and have tested various cases I can see, but please let me know if I've missed anything and I'll address it.

The new command line parameter is -presalphacov and takes the alpha test reference which will be used at runtime. For example: texconv -presalphacov 0.8 cat.jpg


Texconv/texconv.cpp
```diff
@@ -184,6 +185,7 @@ const SValue g_pOptions[] =
     { L"c",             OPT_COLORKEY },
     { L"tonemap",       OPT_TONEMAP },
     { L"x2bias",        OPT_X2_BIAS },
+    { L"presalphacov",  OPT_PRESERVE_ALPHA_COVERAGE },
```
walbourn: Maybe ``keepcoverage`` would be easier to parse?

DirectXTex/DirectXTex.h
```diff
@@ -528,6 +528,13 @@ namespace DirectX
         // levels of '0' indicates a full mipchain, otherwise is generates that number of total levels (including the source base image)
         // Defaults to Fant filtering which is equivalent to a box filter
 
+    HRESULT __cdecl CalculateAlphaCoverage(
```
walbourn: I was wondering why you made ``CalculateAlphaCoverage`` and ``ScaleMipMapsAlphaForCoverage`` distinct public functions instead of having just a single function. Based on the usage in ``texconv`` you just call one then the other...

Texconv/texconv.cpp
jmuulian: Fair point, I couldn't decide on a good command name. "keepcoverage" would work fine.

DirectXTex/DirectXTex.h
jmuulian: No particular reason, other than that's how I structured the problem and solution in my mind. You're right that they're only called in sequence and it's unlikely one would be needed without the other, so I can merge CalculateAlphaCoverage into ScaleMipMapsAlphaForCoverage to clean it up a bit.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -169,6 +169,110 @@ namespace
     }
 
 
+    void GenerateAlphaCoverageConvolutionVectors(
+        _In_ size_t N,
+        _Out_ XMVECTOR* vectors)
```
walbourn: Note that the SAL for this should be ``_Out_writes_(N*N)``

DirectXTex/DirectXTex.h
```diff
@@ -528,11 +528,9 @@ namespace DirectX
         // levels of '0' indicates a full mipchain, otherwise is generates that number of total levels (including the source base image)
         // Defaults to Fant filtering which is equivalent to a box filter
 
-    HRESULT __cdecl CalculateAlphaCoverage(
-        _In_ const Image& srcImage, _In_ float alphaReference, _In_ float alphaScale, _Out_ float& coverage);
     HRESULT __cdecl ScaleMipMapsAlphaForCoverage(
```
walbourn: Looks good being a single function. I do wonder if the name is obvious about what it does, but I can't think of a better one...

DirectXTex/DirectXTexMipmaps.cpp
jmuulian: Good spot. Want me to fix this and update the PR?

walbourn: Probably simplest for you to fix, yes. Thanks!

jmuulian: Done!

walbourn:
Thanks for the contribution. I made a merge fix for my texconv update for ``-inverty``.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -116,6 +116,211 @@ namespace
...
+            const XMVECTOR* pRow0 = row0.get();
+            const XMVECTOR* pRow1 = row1.get();
+            for (size_t x = 0; x < srcImage.width - 1; ++x)
+            {
+                // [0]=(x+0, y+0), [1]=(x+0, y+1), [2]=(x+1, y+0), [3]=(x+1, y+1)
+                alpha.r[0] = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*pRow0), scale));
+                alpha.r[1] = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*pRow1), scale));
+                alpha.r[2] = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*(pRow0++)), scale));
+                alpha.r[3] = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*(pRow1++)), scale));
+                XMVECTOR v = XMVectorSet(XMVectorGetX(alpha.r[0]), XMVectorGetX(alpha.r[1]), XMVectorGetX(alpha.r[2]), XMVectorGetX(alpha.r[3]));
```
walbourn: This converts to scalar and back. Keeping it vector you can do the same rearrangement as:

                XMVECTOR v1 = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*pRow0), scale));
                XMVECTOR v2 = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*pRow1), scale));
                XMVECTOR v3 = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*(pRow0++)), scale));
                XMVECTOR v4 = XMVectorSaturate(XMVectorMultiply(XMVectorSplatW(*(pRow1++)), scale));

                v1 = XMVectorMergeXY(v1, v2); // [v1.x v2.x --- ---]
                v3 = XMVectorMergeXY(v3, v4); // [v3.x v4.x --- ---]

                XMVECTOR v = XMVectorPermute<0, 1, 4, 5>(v1, v3); // [v1.x v2.x v3.x v4.x]

I'm adding this optimization to the main branch...

walbourn: On SSE, ``XMVectorPermute<0, 1, 4, 5>(v1, v3)`` is a single instruciton ``_mm_movelh_ps``

walbourn:
Some code review feedback is [here](https://github.com/Microsoft/DirectXTex/commit/cded42acfc7ee84f1f46da47f393f40b5e64c72c)

walbourn:
And the docs are on the [wiki](https://github.com/Microsoft/DirectXTex/wiki/ScaleMipMapsAlphaForCoverage)

PR #83, opened 2018-03-27
Optimize BC7 parity bit calculation

elasota:
Major speed optimization, reduces encode times with -bcmax in particular by more than half.

The current encoder solves parity bits by trying the block with every possible P-bit combination.  Most of these combinations are redundant in partitioned modes because the P-bits are associated with independent subsets, so it can loop far fewer times by solving each subset separately instead of the entire block.

Mode 0 is the most affected, its loop count is reduced by a factor of 16.

### DIFF
--- a/DirectXTex/Shaders/BC7Encode.hlsl
+++ b/DirectXTex/Shaders/BC7Encode.hlsl
@@ -680,16 +680,16 @@ void TryMode137CS( uint GI : SV_GroupIndex, uint3 groupID : SV_GroupID ) // mode
         if (1 == g_mode_id)
         {
             // in mode 1, there is only one p bit per subset
-            max_p = 4;
+            max_p = 2;
         }
         else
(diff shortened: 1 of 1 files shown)


walbourn:
I'll be taking a look at this this week. I'm setting up a test run to validate quality/time between the two versions.

elasota:
This one should be producing identical output, you should be able to just compress a bunch of test images and compare their checksums.

walbourn:
I'll take this PR once I have my BC validator test done... I've been focused on VS 2017 (15.7 update) and Windows 10 SDK (17134) the past month, so I'm getting back to this...

walbourn:
Thanks again. Fixed in [this commit](https://github.com/Microsoft/DirectXTex/commit/34c6849d74c07e3e61d26203402bd478a1cbfd43)

PR #86, opened 2018-04-05
New BC7 CPU encoder

elasota:
New BC7 CPU encoder, significantly improves speed and quality.

Uses a mix of existing techniques (PCA + k-means clustering/LLS refinement) and a new one that tries to trade index precision for endpoint precision, and processes 8 blocks at once using SSE2.

Based on benchmarks against every other BC7 compressor that I'm aware of (both existing DirectXTex encoders, FasTC, Intel ISPC Texture Compressor, NVTT, Compressonator, bc7_gpu), it's roughly tied for best quality with NVTT and Compressonator, but is about 35x as fast as both of those and 10x as fast as the existing BC7 CPU encoder.

### DIFF
--- a/DirectXTex/BC7Parallel.cpp
+++ b/DirectXTex/BC7Parallel.cpp
@@ -1236,8 +1236,11 @@ namespace
 
         static void CompressEndpoints3(MInt16 ep[2][4], uint16_t p[2])
         {
-            for (int j = 0; j < 2; j++)
-                QuantizeP(ep[j], 7, p[j], 3);
+			for (int j = 0; j < 2; j++)
+			{
+				QuantizeP(ep[j], 7, p[j], 3);
+				ep[j][3] = ParallelMath::MakeUInt16(255);
+			}
         }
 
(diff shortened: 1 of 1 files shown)


walbourn:
Thanks. I appreciate the work here and the contribution. I do need to take some time to review it in detail, and I'm a little concerned about the comingling of various licenses here (i.e. it's just not MIT).

elasota:
The Squish license and the Convection license are both MIT, so it's all one license.

walbourn:
What exactly came from libsquish?

elasota:
Just the initial endpoint selector code (a.k.a. the EndpointSelectorRGB and EndpointSelectorRGBA).

elasota:
Replaced the Squish dependency with a fresh implementation and fixed channel weighting in the process, merged in an important speed optimization.  I think that should be it for any significant changes until review.

Guekka:
Is this PR still in work? It would be very useful on Linux

Issue #92, opened 2018-04-23
Error increases when BC_FLAGS_USE_3SUBSETS flag is specified


richgel999:
When I call DirectX::D3DXEncodeBC7() with DirectX::BC_FLAGS_USE_3SUBSETS, the overall output error increases. This is incorrect: the overall error should decrease. As far as I can tell this is a bug and could be indicative of a deeper issue with this encoder. By comparison, ispc_texcomp does not exhibit this behavior:

[More info](http://richg42.blogspot.com/2018/04/directxtex-bc7-3-subset-wierdness.html)

richgel999:
Ok, the bug is caused by D3DX_BC7::Refine() not correctly computing the block error in a way that factors in the actual coded pbits. It computes the error using the endpoints before the LSB's are slammed to the pbits, causing the encoder to choose blocks that actually increase error. This will wreck quality in many modes.

I've attached a simple fix which introduces a new helper function D3DX_BC7::FixEndpointPBits(), which Refine() calls after it computes new endpoints. It could be easily improved further for more gains (it doesn't round the component values correctly factoring in the pbits), but it's definitely far better than what's there which is broken. I don't use git often but if you want I can submit a PR.

With this single fix applied the encoding quality across my test corpus increased from 44.25 to 46.02 with 3 subsets enabled, and 45.64 with 3 subsets disabled. (My encoder and ispc_texcomp get >46.5 dB with 3 subsets enabled, so this isn't that bad now.) This fix boosts the encoder's default (2 subset modes only) quality above what it was previously capable of using 3 subset modes.

[BC6HBC7.zip](https://github.com/Microsoft/DirectXTex/files/1940466/BC6HBC7.zip)

walbourn:
Please submit it as a pull request. Thanks!

walbourn:
See [this pull request](https://github.com/Microsoft/DirectXTex/pull/93)

walbourn:
Fixed this specific issue in [this commit](https://github.com/Microsoft/DirectXTex/commit/3931c8866f43e5535af9133403ddb0a08dc9fba4)

PR #93, opened 2018-04-24
pbit related quality fix for BC7 texture encoding

richgel999:
When I call DirectX::D3DXEncodeBC7() with DirectX::BC_FLAGS_USE_3SUBSETS, the overall output error increases across the kodim test corpus. This is incorrect: the overall error should decrease. Other BC7 encoders don't behave in this way, so I investigated further. I modified D3DX_BC7::Refine() to call D3DXDecodeBC7() on the encoded block, computed the actual error, and compared that vs. the error computed by Refine(). They differed, which is incorrect.

The bug is caused by D3DX_BC7::Refine() not correctly computing the block error in a way that factors in the actual coded pbits. It computes the error using the endpoints before the LSB's are slammed to the pbits, causing the encoder to choose blocks that actually increase error. This will wreck quality in pbit modes.

The fix introduces a new helper function D3DX_BC7::FixEndpointPBits(), which Refine() calls after it computes new endpoints. It could be easily improved further for more gains (it doesn't round the component values correctly factoring in the pbits), but it's a massive improvement.

With this fix applied the encoding quality across my test corpus increased from 44.25 to 46.02 with 3 subsets enabled, and 45.64 with 3 subsets disabled. Tested with opaque and transparent textures.

### DIFF
--- a/DirectXTex/BC.h
+++ b/DirectXTex/BC.h
@@ -36,6 +36,7 @@ enum BC_FLAGS
     BC_FLAGS_UNIFORM            = 0x40000,  // By default, uses perceptual weighting for BC1-3; this flag makes it a uniform weighting
     BC_FLAGS_USE_3SUBSETS       = 0x80000,  // By default, BC7 skips mode 0 & 2; this flag adds those modes back
     BC_FLAGS_FORCE_BC7_MODE6    = 0x100000, // BC7 should only use mode 6; skip other modes
+    BC_FLAGS_FASTER_BC7         = 0x200000, // BC7 should try much less harder, but still use most modes - ~10x faster
 };
 
 //-------------------------------------------------------------------------------------
--- a/DirectXTex/BC6HBC7.cpp
+++ b/DirectXTex/BC6HBC7.cpp
@@ -854,7 +854,7 @@ namespace
             _In_reads_(NUM_PIXELS_PER_BLOCK) const size_t aIndex[],
             _In_reads_(NUM_PIXELS_PER_BLOCK) const size_t aIndex2[]);
        float Refine(_In_ const EncodeParams* pEP, _In_ size_t uShape, _In_ size_t uRotation, _In_ size_t uIndexMode);
-       void FixEndpointPBits(_In_ const EncodeParams* pEP, _In_ const LDREndPntPair *pOrigEndpoints, _Out_writes_(3) LDREndPntPair *pFixedEndpoints);
+       void FixEndpointPBits(_In_ const EncodeParams* pEP, _In_reads_(BC7_MAX_REGIONS) const LDREndPntPair *pOrigEndpoints, _Out_writes_(BC7_MAX_REGIONS) LDREndPntPair *pFixedEndpoints);
 
         float MapColors(_In_ const EncodeParams* pEP, _In_reads_(np) const LDRColorA aColors[], _In_ size_t np, _In_ size_t uIndexMode,
             _In_ const LDREndPntPair& endPts, _In_ float fMinErr) const;
--- a/DirectXTex/BC6HBC7.cpp
+++ b/DirectXTex/BC6HBC7.cpp
@@ -787,6 +787,7 @@ namespace
         struct EncodeParams
         {
             uint8_t uMode;
+			bool bFaster;
             LDREndPntPair aEndPts[BC7_MAX_SHAPES][BC7_MAX_REGIONS];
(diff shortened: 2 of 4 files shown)


walbourn:
Thanks. I'm working on a validator test to get better regression coverage when taking BC changes. Once I'm able to run a baseline I can try your change and verify it works as advertised.

richgel999:
Yes, independent validation is really necessary and I'll wait on submitting any more improvements until you have this going.

I have another set of simple changes that introduces a new flag (BC_FLAGS_FASTER_BC7) that improves encoding speed by ~10x. Quality (without 3 subsets enabled) across my corpus is reduced by 1 dB (from 45.64 to 44.65), but it's still stronger than the original encoder with the pbit bug (which was 44.4 dB). Without the flag the encoder works as before, and it's combinable with the other flags (3 subsets and quick). The changes are very simple because they just leverage the existing encoder and add several heuristics used by other encoders. 

The first fix disables mode 7 on completely opaque blocks. Mode 7 isn't useful because the other modes are just as strong for opaque cases. There's no quality loss from this change, but it's noticeably faster (12429 vs. 9756 secs).

The other improvements: Only refine a single partition for each mode, use bounding box approximation vs. full PCA in OptimizeRGB/OptimizeRGBA when estimating partitions, use mode 1's estimated partition for mode 3, and disable the deeper exhaustive refinement. 

										CPU Time	 RGB PSNR    	
No flags:
Orig/unmodified:							12505.338093 45.638377
After my changes, remark out m7 for opaque       12429.388720 45.638377
After my changes, skip m7 for opaque:  	         9755.917509  45.638270

BC_FLAGS_FASTER_BC7:
Only refine a single partition:  			 	3358.698415  45.360886
Estimated partition selection:   			 	2884.865429  45.037020
Disable less useful modes for opaque:   	  	2821.762252  45.017103
Same partition for 1,3, less refinement:	 	        1189.376734  44.648705

BC_FLAGS_FORCE_BC7_MODE6:
BC_FLAGS_FORCE_BC7_MODE6					   	471.434405   41.863792
BC_FLAGS_FORCE_BC7_MODE6|BC_FLAGS_FASTER_BC7   	188.634432   41.070086

richgel999:
I've checked in the code for the new flag:
https://github.com/Microsoft/DirectXTex/pull/93/commits/d3f1d3c8f230ac2d2c321930c5bcf12d3750cc2f

I've tested with both opaque and complex transparent textures.

elasota:
Should probably fix the indentation formatting.  The existing code uses spaces, but these changes use hard tabs.

richgel999:
Ok - I'll fix this within a day or two.

walbourn:
Sorry I thought this was the pull request for that issue. If not, please submit it as a distinct PR. Thanks!

richgel999:
BTW - ImageMagick can compare images and output PSNR:
http://www.imagemagick.org/Usage/compare/#compare

I like using tools like this because it's a good independent comparison point. My PSNR's are compatible with this tool.

walbourn:
I appreciate the contribution, but this pull request seems to merge a lot of distinct optimizations into one big change so I'm having trouble validating each part of it...

I'm going to try to tease apart the different fixes here, in particular the mode 7/opaque skip case and the original fix for [this issue](https://github.com/Microsoft/DirectXTex/issues/92) which I want to try out independent of your new suggestion "FASTER" flag...

walbourn:
I've taken the fix with ``FixEndpointPBits`` in [this commit](https://github.com/Microsoft/DirectXTex/commit/3931c8866f43e5535af9133403ddb0a08dc9fba4), so this PR now fails due to conflicts...

walbourn:
I've taken the fix to skip mode 7 for opaque blocks in [this commit](https://github.com/Microsoft/DirectXTex/commit/ba30da0945c266fa8e765ba812d8d90b9d29796a).

If you'd like to clean up the conflicts and update the PR for just the 'faster' mode, I can look at taking those changes.

Thanks again!

richgel999:
Thanks Chuck. I've been busy on our universal format, but once I go back to BC7 I can fix up the changelist for faster mode.

walbourn:
Have you had a chance to look at the conflicts?

walbourn:
Any update?

richgel999:
Hi - Sorry, we had to ship our universal codec. It looks like you've already merged all of the key bug fixes. The p-bit thing was the most important.
Best regards,
-Rich

PR #100, opened 2018-05-13
Add FP16 NaN/infinity handling flags

elasota:
Adds 2 flags for more control over special floating point values when converting to FP16.  May be useful for applications that depend on NaN/infinity for special handling, like using NaN as a transparency sentinel.

TEX_FILTER_FLOAT_SATURATE_TO_INF saturates out-of-range values to infinity, including existing infinities.  TEX_FILTER_FLOAT_KEEP_NANS keeps NaNs instead of forcing them to zero.

Default behavior is convert NaN to zero and saturate to max range.


walbourn:
I'd have implemented these as part of the ``Convert`` special-case path rather than ``LoadScanline``/``StoreScanline`` since it's specific to conversion codepaths.

Since it's specific to FP32 to FP16 conversions, maybe name the flags: ``TEX_FILTER_FLOAT16_*``

walbourn:
I totally get what you are going for here, I'm just wondering about some other impacts this might have elsewhere. I'll think about it a bit more...

walbourn:
Revisiting this issue part of the problem I think is the functionality offered in CVT16/F16C. These instructions let you control rounding, but don't have any ability to control NANs and/or DENORMS. I of course can implement it in software, but it will slow it down a bit.

As such, I think the default behavior needs to be "use NANs and DENORMs" and provide explicit flags for flushing DENORMs (which would need to be implemented for both F16 and F32 to make sense) and preferring INFs to specials for conversions. 

I have some DirectXMath [bugs](https://github.com/microsoft/DirectXMath/issues/110) in the handling of NANs. and DENORMS to deal with first, and when I wrap those up I'll revisit this PR.

DirectXTex/DirectXTexConvert.cpp
```diff
@@ -1659,9 +1659,7 @@ bool DirectX::Internal::StoreScanline(
             for (size_t icount = 0; icount < (size - sizeof(XMHALF4) + 1); icount += sizeof(XMHALF4))
             {
                 if (sPtr >= ePtr) break;
-                XMVECTOR v = *sPtr++;
-                v = XMVectorClamp(v, g_HalfMin, g_HalfMax);
-                XMStoreHalf4(dPtr++, v);
+                XMStoreHalf4(dPtr++, *sPtr++);
```
walbourn, line 1662: I'm a little nervous about removing the clamp behavior here, but it's probably the best choice.

The main places where the clamping is now 'missing' is:

* BC Decompression - this shouldn't ever have specials
* TransformImage - The pixel function used needs to clamp if they care
* ConvertToRGBA32 helper - not using float16 formats as the target
* Custom resize, Custom mip generation, CopyRectange, NormalMap generation, PM Alpha conversion - If the input has specials, the output will have them.

I think the ``ConvertFromR32G32B32A32`` helpers needs an explicit clamp added for the float16 cases. This is the primarily used for WIC interactions where WIC doesn't support the format.

DirectXTex/DirectXTexConvert.cpp
```diff
@@ -4441,6 +4458,12 @@ namespace
             return false;
         }
 
+        if (filter & (TEX_FILTER_FLOAT16_SATURATE_TO_INF | TEX_FILTER_FLOAT16_KEEP_NANS))
```
walbourn, line 4461: Before I can sign off here, I really need to see what WIC does. Does it preserve specials or always clamp them?

DirectXTex/DirectXTexConvert.cpp
elasota: >     * BC Decompression - this shouldn't ever have specials

BC6H signed can have -INF, even though it's not recommended.

PR #103, opened 2018-05-31
DirectTexImage::IsAlphaAllOpaque() changed threshold to handle a case when opacity of pixel defined with color (254,254,254)

apatynko:
the threshold was changed from { 0.99f, 0.99f, 0.99f, 0.99f } to { 0.997f, 0.997f, 0.997f, 0.997f }
to handle a case when opacity of pixel defined with color (254,254,254)
 in this case alpha = { 0.996078491f, 0.996078491f, 0.996078491f, 0.996078491f }  
All others color values from 0 to 253 works fine with threshold=0.99f, only "254" has exception.

### DIFF
--- a/DirectXTex/DirectXTexImage.cpp
+++ b/DirectXTex/DirectXTexImage.cpp
@@ -764,7 +764,7 @@ bool ScratchImage::IsAlphaAllOpaque() const
         if (!scanline)
             return false;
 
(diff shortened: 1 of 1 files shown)


walbourn:
I added an explicit test to cover this scenario and it looks like 253 and 254 both fail without this fix.

Thanks!

PR #106, opened 2018-06-12
Fix compiler errors (Clang 6)

walbourn:
Fix for ``InitOnceExecuteOnce`` failing to build with clang x86.

Also various code review cleanup based on clang warnings.

### DIFF
--- a/DirectXTex/BC.cpp
+++ b/DirectXTex/BC.cpp
@@ -9,7 +9,7 @@
 // http://go.microsoft.com/fwlink/?LinkId=248926
 //-------------------------------------------------------------------------------------
 
-#include "directxtexp.h"
+#include "DirectXTexp.h"
 
 // Experiemental encoding variants, not enabled by default
 //#define COLOR_WEIGHTS
@@ -35,9 +35,9 @@ namespace
     //-------------------------------------------------------------------------------------
     inline void Decode565(_Out_ HDRColorA *pColor, _In_ const uint16_t w565)
     {
-        pColor->r = (float)((w565 >> 11) & 31) * (1.0f / 31.0f);
-        pColor->g = (float)((w565 >> 5) & 63) * (1.0f / 63.0f);
-        pColor->b = (float)((w565 >> 0) & 31) * (1.0f / 31.0f);
+        pColor->r = static_cast<float>((w565 >> 11) & 31) * (1.0f / 31.0f);
+        pColor->g = static_cast<float>((w565 >> 5) & 63) * (1.0f / 63.0f);
+        pColor->b = static_cast<float>((w565 >> 0) & 31) * (1.0f / 31.0f);
         pColor->a = 1.0f;
     }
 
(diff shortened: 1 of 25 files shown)


DirectXTex/BC.cpp
```diff
@@ -35,9 +35,9 @@ namespace
     //-------------------------------------------------------------------------------------
     inline void Decode565(_Out_ HDRColorA *pColor, _In_ const uint16_t w565)
     {
-        pColor->r = (float)((w565 >> 11) & 31) * (1.0f / 31.0f);
-        pColor->g = (float)((w565 >> 5) & 63) * (1.0f / 63.0f);
-        pColor->b = (float)((w565 >> 0) & 31) * (1.0f / 31.0f);
+        pColor->r = static_cast<float>((w565 >> 11) & 31) * (1.0f / 31.0f);
```
walbourn: ``-Wold-school-casts`` in clang found a lot of these lingering in the code, so I fixed them.

DirectXTex/BC.cpp
```diff
@@ -9,7 +9,7 @@
 // http://go.microsoft.com/fwlink/?LinkId=248926
 //-------------------------------------------------------------------------------------
 
-#include "directxtexp.h"
```
walbourn: For testing I put the ``#pragma clang diagnostic ignore`` suppressions into the pch here, but those suppressions weren't in place before it was included, so you got a bunch of ``-Wnonportable-include-path`` warnings, so I changed these to make it less noisy...

DirectXTex/BC6HBC7.cpp
```diff
@@ -605,7 +600,7 @@ namespace
             {
                 size_t uFirstIndexBits = 8 - uBase;
                 size_t uNextIndexBits = uNumBits - uFirstIndexBits;
-                ret = (m_uBits[uIndex] >> uBase) | ((m_uBits[uIndex + 1] & ((1 << uNextIndexBits) - 1)) << uFirstIndexBits);
+                ret = static_cast<uint8_t>((unsigned(m_uBits[uIndex]) >> uBase) | ((unsigned(m_uBits[uIndex + 1]) & ((1u << uNextIndexBits) - 1)) << uFirstIndexBits));
```
walbourn: ``-Wconversion`` warnings happened in a lot of places because in most cases the bit-wise operators promote to an ``int`` even if they smaller size is unsigned.

DirectXTex/BCDirectCompute.cpp
```diff
@@ -608,4 +608,4 @@ HRESULT GPUCompressBC::Compress(const Image& srcImage, const Image& destImage)
     }
 
     return hr;
-}
\ No newline at end of file
+}
```
walbourn: ``-Wnewline-eof`` which would also be gripped about with ``-Wc++98-compat-pedantic``

DirectXTex/DirectXTexD3D11.cpp
```diff
@@ -763,7 +763,7 @@ HRESULT DirectX::CaptureTexture(
             desc.CPUAccessFlags = D3D11_CPU_ACCESS_READ;
             desc.Usage = D3D11_USAGE_STAGING;
 
-            hr = pDevice->CreateTexture1D(&desc, 0, pStaging.GetAddressOf());
+            hr = pDevice->CreateTexture1D(&desc, nullptr, pStaging.GetAddressOf());
```
walbourn: I had fixed most of the ``-Wzero-as-null-pointer-constant`` in WIC usage already, but missed a few...

PR #107, opened 2018-06-28
fix EncodeBC1 floating point exception

poulpi:
Floating point exception sometimes happens when calling HDRColorALerp() in EncodeBC1() which lerp the uninitialized (and unused) alpha value.

### DIFF
--- a/DirectXTex/BC.cpp
+++ b/DirectXTex/BC.cpp
@@ -569,14 +569,14 @@ namespace
         {
             pSteps = pSteps3;
 
-            HDRColorALerp(&Step[2], &Step[0], &Step[1], 0.5f);
+            HDRColorALerpRGB(&Step[2], &Step[0], &Step[1], 0.5f);
         }
         else
         {
             pSteps = pSteps4;
 
-            HDRColorALerp(&Step[2], &Step[0], &Step[1], 1.0f / 3.0f);
-            HDRColorALerp(&Step[3], &Step[0], &Step[1], 2.0f / 3.0f);
+            HDRColorALerpRGB(&Step[2], &Step[0], &Step[1], 1.0f / 3.0f);
+            HDRColorALerpRGB(&Step[3], &Step[0], &Step[1], 2.0f / 3.0f);
         }
 
         // Calculate color direction
--- a/DirectXTex/BC.h
+++ b/DirectXTex/BC.h
@@ -141,6 +141,14 @@ inline HDRColorA* HDRColorALerp(_Out_ HDRColorA *pOut, _In_ const HDRColorA *pC1
     return pOut;
 }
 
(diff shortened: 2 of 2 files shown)


walbourn:
Can you provide a bit more information about the underlying conditions of why you felt this was a problem?

FP-specials are not enabled by default, and turning them on in code outside your control is problematic to begin with.

walbourn:
A more direct fix might also be to just initialize the ``HDRColorA`` struct so that ``.a`` is 0...

walbourn:
Please see if the changes in [this commit](https://github.com/Microsoft/DirectXTex/commit/5e675e01a2ab95e9351e59ca97291e247319de96) resolve your issue... If not, feel free to reopen.

poulpi:
We are using FP exceptions to help us tracks bugs, and we use DirectXTex at runtime as with need texture compression on the fly.

The commit you proposed do not fix it. The problematic value is created l494 before the OptimizeRGB(), and crash in HDRColorALerp() l578.

Initializing the rgba values in the HDRColorA class constructor may be the best fix ?

walbourn:
Can you give it one more try with  [this commit](https://github.com/Microsoft/DirectXTex/commit/5004dd770cf4cc800ccdd9889e05720b02b55c57)? I found another case I missed.

BTW, can you provide the snippet of code you are using to enable FPs exceptions for your debug builds? I can see about adding that to my [DirectXTex test suite](https://github.com/walbourn/directxtextest).

poulpi:
The commit works, thanks you.

Here the code we use to set or unset floating points exceptions:
```
#define FLOAT_EXCEPTION_FLAGS (_EM_ZERODIVIDE | _EM_OVERFLOW | _EM_INVALID | _EM_DENORMAL)
unsigned int x86_cw;
```
set:
```
#if defined (_M_IX86)
	_clearfp(); 
	__control87_2( _EM_UNDERFLOW|_EM_INEXACT|_EM_DENORMAL|_EM_ZERODIVIDE, _MCW_EM, &x86_cw, NULL );
#elif defined(_M_X64)
	_controlfp_s(&x86_cw, 0, FLOAT_EXCEPTION_FLAGS);
#endif
```
unset:
```
#if defined (_M_IX86)
	_clearfp(); 
	__control87_2( _MCW_EM, _MCW_EM, &x86_cw, NULL );
#elif defined(_M_X64)
	_controlfp_s(&x86_cw, FLOAT_EXCEPTION_FLAGS, FLOAT_EXCEPTION_FLAGS);
#endif
```

walbourn:
Thanks.

Note that I'm not sure your ``_M_IX86`` case does what you think it does. Unless you use ``/arch:IA32``, the Visual C++ 32-bit compiler is going to use SSE/SSE2 and not x87 for floating-point computations, and it appears from the code above you are not setting the ``SSE/SSE`` control word in x86 processes. You should probably just use ``_controlfp_s`` for all platforms.

walbourn:
Updated test suite in [this commit](https://github.com/walbourn/directxtextest/commit/ae4340739bd60e7924b20095775adafe554fc3a5)

PR #108, opened 2018-06-29
Update GPU encoder to require DirectCompute 5

walbourn:
The origin BC6HBC7EncoderCS project supported DirectCompute 4.x on Feature Level 10.x cards due to the limited number of FL 11 cards at the time.

This change updates the shaders to use ``cs_5_0`` profiles.


walbourn:
Testing is getting mixed results in terms of speed & quality when using CS 5.0 vs. CS 4.0. Need to do some more investigation...

walbourn:
Bad timings are coming from a non-consumer brand of video card, so it may just be driver issues...

DirectXTex/BCDirectCompute.cpp
```diff
@@ -20,13 +20,29 @@ using Microsoft::WRL::ComPtr;
...
+        #include "Shaders\Compiled\BC7Encode_EncodeBlockCS.inc"
+        #include "Shaders\Compiled\BC7Encode_TryMode02CS.inc"
+        #include "Shaders\Compiled\BC7Encode_TryMode137CS.inc"
+        #include "Shaders\Compiled\BC7Encode_TryMode456CS.inc"
+        #include "Shaders\Compiled\BC6HEncode_EncodeBlockCS.inc"
+        #include "Shaders\Compiled\BC6HEncode_TryModeG10CS.inc"
+        #include "Shaders\Compiled\BC6HEncode_TryModeLE10CS.inc"
+    }
+
+#if !defined(_XBOX_ONE) || !defined(_TITLE)
```
walbourn: This logic is just to keep the Xbox One static library a little smaller since it will always use the CS 5.0 version.

DirectXTex/Shaders/BC6HEncode.hlsl
```diff
@@ -979,14 +979,15 @@ void EncodeBlockCS(uint GI : SV_GroupIndex, uint3 groupID : SV_GroupID)
     }
 }
 
-uint float2half1( float f )
+#ifdef EMULATE_F16C
```
walbourn: The original implementation emulated the ``f32tof16`` intrinsics directly. While the HLSL compiler will emulate this for you on SM 4.x, it turns out the emulation is slower than the original code on 4.x targets. It is faster to use the intrinsic on SM 5.0 targets.

Therefore, the logic here in this change is to use the original emulation code for SM 4.0 and the intrinsics for SM 5.0.

walbourn:
It looks like this change is basically the same for BC7, but the BC6H SM 5.0 version is smaller code and faster generally.

walbourn:
Refreshed this PR for the latest changes to main.

walbourn:
Verified with modern drivers that the two versions are basically the same quality and performance, with BC6 version being a bit faster on CS 5.0 due to the use of the f32tof16 / f16tof32 intrinsics.

PR #109, opened 2018-07-02
Fix out-of-bounds read

elasota:
Fix OOB memory read.

To reproduce, download this file:
https://github.com/knarkowicz/GPURealTimeBC6H/blob/master/bin/yucca.dds

Then run:
texconv.exe -nogpu -f BC6H_SF16 -m 1 -y yucca.dds

At some points, fDot will be -1.2, causing iStep to underflow, causing a crash.


walbourn:
Great catch. Thanks for the test case too!

PR #112, opened 2018-07-28
Workaround when using WIC for F32 images that exceed 32-bits

walbourn:
The WIC interface uses a ``UINT`` (32-bit) for the total size of an image. This works for many cases, but fails for large images when converting to F32 (i.e. 16384 x 16384 converted to a 128bppFloat exceeds 32-bits).

This pull request uses 64bppHalf instead if the image is too large to fit into 32-bits when doing WIC operations with ``FlipRotate``, and uses custom filtering for ``Resize`` and ``GenerateMipMaps``


DirectXTex/DirectXTexConvert.cpp
```diff
@@ -2627,6 +2627,178 @@ HRESULT DirectX::_ConvertFromR32G32B32A32(
...
+    const uint8_t *pSrc = srcImage.pixels;
+    for (size_t h = 0; h < srcImage.height; ++h)
+    {
+        if (!_LoadScanline(scanline.get(), srcImage.width, pSrc, srcImage.rowPitch, srcImage.format))
+        {
+            image.Release();
+            return E_FAIL;
+        }
+
+        if (!_StoreScanline(pDest, img->rowPitch, DXGI_FORMAT_R16G16B16A16_FLOAT, scanline.get(), srcImage.width))
```
walbourn: I could use DirectXMath's ``XMConvertFloatToHalfStream`` here, but I want to actually try out some test cases first. The main reason to use the ``_StoreScanline`` solution is that it explicitly clamps the values into the float16 min/max range, while the DirectXMath conversion saturates to infinity.

walbourn: Problem cases are very large images in the following formats (non-WIC): ``DXGI_FORMAT_R32G32B32A32_UINT``, ``DXGI_FORMAT_R32G32B32A32_SINT``, ``DXGI_FORMAT_R32G32B32_UINT``, and ``DXGI_FORMAT_R32G32B32_SINT`` which exceed the range of a float16.

DirectXTex/DirectXTexConvert.cpp
```diff
@@ -2627,6 +2627,178 @@ HRESULT DirectX::_ConvertFromR32G32B32A32(
...
+    const uint8_t *pSrc = srcImage.pixels;
+    for (size_t h = 0; h < srcImage.height; ++h)
+    {
+        if (!_LoadScanline(scanline.get(), srcImage.width, pSrc, srcImage.rowPitch, srcImage.format))
+        {
+            image.Release();
+            return E_FAIL;
+        }
+
+        XMConvertFloatToHalfStream(
```
walbourn: This function doesn't clamp, it saturates to infinity. We are already in 'doing our best' territory here with such large images, so I think it's probably better to have INFs in the results than sharply clamped values when they exceed the range of a float16.

DirectXTex/DirectXTexFlipRotate.cpp
```diff
@@ -206,7 +251,16 @@ HRESULT DirectX::FlipRotate(
     else
     {
         // Case 2: Source format is not supported by WIC, so we have to convert, flip/rotate, and convert back
-        hr = PerformFlipRotateViaF32(srcImage, flags, *rimage);
```
walbourn: So direct attempts to flip/rotate, resize, or generate mips on a 16834 x 16834 ``DXGI_FORMAT_R32B32G32A32_FLOAT`` image using WIC is going to continue to fail even with these changes as WIC just can't handle that size. The purpose of this fix is to make *other* formats work that don't have a direct DXGI-format equivalent in WIC, which when converted to ``DXGI_FORMAT_R32B32G32A32_FLOAT`` exceed a 32-bit size.

walbourn:
Note that I currently can't read or write a 16384 x 16834 DDS file containing ``D3DFMT_A32B32G32R32F`` / ``DXGI_FORMAT_R32G32B32A32_FLOAT`` data as the current reader rejects files larger than 4GB. I could in theory load it in 64-bit native platforms, but woudl have to do a bit of work in both DirectXTex and DirectX Tool Kit because ``ReadFile`` only reads 32-bits of data at a time...

walbourn:
Using ``GUID_WICPixelFormat64bppRGBAHalf`` instead of ``GUID_WICPixelFormat128bppRGBAFloat`` as an intermediate WIC surface for working with non-WIC formats when the image exceeds a 32-bit integer in size works fine for ``IWICBitmapFlipRotator``. It returns the data in  ``GUID_WICPixelFormat64bppRGBAHalf``. That resolves the problem of  using ``FlipRotate`` on large images.

walbourn:
This pull request improves the handling of large files when using the x64 native version of the library or ``texconv``, but there are still a number of limitations due to the fact that WIC only supports up to 4 GB images. I still defer to using WIC by default where the format is supported, and those cases will likely fail when the ``IWICScalar`` can't return the result in a small enough image.

You can workaround those individual cases with the ``TEX_FILTER_FORCE_NON_WIC`` flag. I've added the ``-nowic`` flag to ``texconv`` / ``texassemble`` to make this scenario a bit easier to deal with when using the command-line tool.

walbourn:
Actually, it looks like on Windows 10 (15063) or older versions of WIC, the ``IWICBitmapScaler`` when given a ``GUID_WICPixelFormat64bppRGBAHalf`` returns the data in ``GUID_WICPixelFormat32bppBGRA`` which really messes it up.

As such, I think for ``Resize`` and ``GenerateMipMaps`` I should just use custom filtering when either the source or the target is going to exceed a UINT as float32

PR #134, opened 2019-04-11
Fix version number in VS2019 solution files

Qartar:
Without this change the Visual Studio Version Selector will try to open these solutions with Visual Studio 2017 and fail to find the v142 toolchain. Version number used is from the 16.0.0 release and used in new solutions created with that version of the IDE. No additional changes were required for individual project files.


walbourn:
Thanks. I'll put this into the other projects I have with VS 2019 projects. I expected the SLN to get upgraded but it didn't.

walbourn:
Note that technically, only the comment line needed updating:

    # Visual Studio 16

Issue #105, opened 2018-06-12
Compilation issues with clang


walbourn:
All Visual C++ architectures and clang x64 all build fine, but clang x86 fails to build the code for ``InitOnceExecuteOnce``:

    InitOnceExecuteOnce(&s_initOnce,
        [](PINIT_ONCE, PVOID, LPVOID *factory) noexcept -> BOOL
    {
    ...
    }

The problem is that clang can't convert the calling convention for the lambda. From [this page](https://clang.llvm.org/docs/MSVCCompatibility.html):

> Lambdas: Mostly complete. Clang is compatible with Microsoft’s implementation of lambdas except for providing overloads for conversion to function pointer for different calling conventions. However, Microsoft’s extension is non-conforming.

walbourn:
If you want a 'clean' build with clang, you need to turn off a bunch of warnings:

    #ifdef __clang__
    #pragma clang diagnostic ignored "-Wc++98-compat"
    #pragma clang diagnostic ignored "-Wc++98-compat-pedantic"
    #pragma clang diagnostic ignored "-Wc++98-compat-local-type-template-args"
    #pragma clang diagnostic ignored "-Wcast-align"
    #pragma clang diagnostic ignored "-Wcovered-switch-default"
    #pragma clang diagnostic ignored "-Wdouble-promotion"
    #pragma clang diagnostic ignored "-Wexit-time-destructors"
    #pragma clang diagnostic ignored "-Wfloat-equal"
    #pragma clang diagnostic ignored "-Wglobal-constructors"
    #pragma clang diagnostic ignored "-Wgnu-anonymous-struct"
    #pragma clang diagnostic ignored "-Wlanguage-extension-token"
    #pragma clang diagnostic ignored "-Wmissing-prototypes"
    #pragma clang diagnostic ignored "-Wnested-anon-types"
    #pragma clang diagnostic ignored "-Wnonportable-include-path"
    #pragma clang diagnostic ignored "-Wnonportable-system-include-path"
    #pragma clang diagnostic ignored "-Wold-style-cast"
    #pragma clang diagnostic ignored "-Wreserved-id-macro"
    #pragma clang diagnostic ignored "-Wshadow-field-in-constructor"
    #pragma clang diagnostic ignored "-Wsign-conversion"
    #pragma clang diagnostic ignored "-Wswitch-enum"
    #pragma clang diagnostic ignored "-Wtautological-type-limit-compare"
[...]

PR #135, opened 2019-04-13
Enable compilation with clang-cl (LLVM on Windows)

Qartar:
Posting for feedback. This is mostly minor fixes for warnings using the LLVM toolchain and one compiler error with the `XMMATRIX` constants. I intend to squash these changes after resolving the remaining `-Wswitch` warnings ("enumeration values not handled in switch", cf. [C4062](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4062)) but want to get some feedback before finishing off the remaining work. I've included new projects here that use the LLVM toolchain for reference but I don't expect to keep those for the merge.


Texdiag/texdiag.cpp
```diff
@@ -1728,8 +1728,6 @@ namespace
 
                         if (bSigned)
                         {
-                        #pragma prefast(push)
```
walbourn: These need to stay for the codebase as I have internal tools that need this.

You can guard these with ``#ifdef _PREFAST_`` which won't be true for clang/LLVM.

    #ifdef _PREFAST_
    #pragma prefast(push)
    #pragma prefast(disable : 26453, "Shift here is never negative")
    #endif

Qartar: Note the change to `SIGN_EXTEND` above, PREFast no longer warns on these lines. FWIW my first attempt in 1b6517e generated 26450 instead ("operation causes overflow at compile time").

walbourn:
Note I've done some cleanups in the past. See [this issue](https://github.com/Microsoft/DirectXTex/issues/105). I feel like at least a few of these warnings just need suppression instead of 'fixing' as they are really just noise and hard to maintain (like ``-Wnonportable-include-path`` and ``-Wnonportable-system-include-path``.

Texdiag/texdiag.cpp
walbourn: Nice. The macro is definitely magic so I didn't have the time to figure that one out :)

Texdiag/texdiag.cpp
```diff
@@ -1368,7 +1368,7 @@ namespace
     }
 
     //--------------------------------------------------------------------------------------
-#define SIGN_EXTEND(x,nb) ((((x)&(1<<((nb)-1)))?((~0)<<(nb)):0)|(x))
+#define SIGN_EXTEND(x,nb) ((((x)&(1<<((nb)-1)))?((~0)^((1<<(nb))-1)):0)|(x))
```
walbourn: This same macro appears in ``BC6HBC7.cpp``, but I don't see that file in the PR...

walbourn:
I'm wiling to take most of these changes, but not hte LLVM ``vcxproj`` or ``sln`` projects as I really don't want to maintain them.

Can you look at why you didn't get BC6BC7.cpp in your review since that has the same SIGN_EXTEND and ``#pragma prefast``?

I do appreciate the work here.

Texdiag/texdiag.cpp
walbourn: I've confirmed that with the ``SIGN_EXTENDED`` change in BC6BC7.cpp, the prefast suppression can also be removed.

Texdiag/texdiag.cpp
Qartar: Good catch. I think the reason it doesn't get flagged by clang-cl in `BC6HBC7.cpp` is because the additive-expression (RHS) is a variable `int(Prec.r))`, not a compile-time constant like it is here. PREFast still flags it though so I'll update it to match.

Qartar:
No problem leaving out the project files, those were more for reference than for merging. I haven't had a chance to look into the `-Wswitch` warnings yet so I'll leave that for another PR. Good news there is that MSVC does have a corresponding warning that can be enabled ([C4062](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4062), off by default).

walbourn:
Accepted the changes here except the vcxproj/slns, plus fixed ``BC6HBC7.cpp`` in [this commit](https://github.com/Microsoft/DirectXTex/commit/86b2bb519440368e38258262d6b3584d62484ec7). You'll probably want to ``git rebase master`` your working branch after syncing master.

Qartar:
Thanks Chuck!

walbourn:
BTW, if you have the time & inclination, a similar clang/LLVM warning pass on [DirectXMesh](https://github.com/Microsoft/DirectXMesh/) would be great. I did a pass on _DirectX Tool Kit_ a while back, so that one is in better shape.

PR #151, opened 2019-10-04
Update TGA codec for TGA 2.0

walbourn:
Three distinct changes in this pull request:

First, the TGA writer now includes the official TGA 2.0 footer in all cases. This is recommended in the spec. It does not impact the ability of older codecs to read the content of the file.

Second, when reading a TGA, if there is a TGA 2.0 extension area present I use the AttributeType to set the alpha mode in the returned metadata--unless the reader already set it to OPAQUE due to format.

Third, if the caller provides the new optional metadata parameter to the TGA writer, it will write the TGA 2.0 extension area. This extension area includes AttributeType derived from the metadata alpha mode, the time stamp, a software ID of "DirectXTex", and a software version of ``DIRECTX_TEX_VERSION``.  If the source format was ``DXGI_FORMAT_*_SRGB`` I also set the gamma value to 2.2.

> I added a ``-tga20`` switch to ``texconv`` which enables the TGA 2.0 extension area writing.

### DIFF
--- a/DirectXTex/DirectXTex.h
+++ b/DirectXTex/DirectXTex.h
@@ -382,8 +382,8 @@ namespace DirectX
         _In_z_ const wchar_t* szFile,
         _Out_opt_ TexMetadata* metadata, _Out_ ScratchImage& image);
 
-    HRESULT __cdecl SaveToTGAMemory(_In_ const Image& image, _Out_ Blob& blob);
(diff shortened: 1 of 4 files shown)


walbourn:
In theory I could use the gamma values of '2.2' to force the return format to ``DXGI_FORMAT_*_SRB`` but I would have to add more parsing to the GetMetadata function. I'm also not sure that this is particular reliable since most TGAs I've encountered don't have the extension area at all.

PR #152, opened 2019-10-10
texconv: added -fixbc4x4 switch

walbourn:
DirectXTex has generalized support for BC image sizes so it can support non-multiple-of-4 image sizes and properly BC them. However, Direct3D itself can only create a BC resource if the top-most miplevel is a multiple-of-4 in both width and height.

As such, some DDS files have been written that can't actually be loaded as textures. For example, a 1x1 BC compressed DDS.  With the new ``-fixbc4x4`` switch, such BC compressed images are rounded up to fix, although this process invalidates any existing mipchain.

### DIFF
--- a/Texconv/texconv.cpp
+++ b/Texconv/texconv.cpp
@@ -107,6 +107,7 @@ enum OPTIONS
     OPT_INVERT_Y,
     OPT_ROTATE_COLOR,
     OPT_PAPER_WHITE_NITS,
+    OPT_BCNONMULT4FIX,
     OPT_MAX
 };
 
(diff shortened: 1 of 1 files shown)


walbourn:
**RE: why cut off mip levels?**

For example, if you have a 9x9 BC image, then rounding it up to nearest multiple-of-4 results in 12x12.

In the original image you had: 9x9 (9 blocks), 4x4 (1 block), 2x2 (1 block), 1x1  (1 block)

In the 'rounded' image it would be: 12x12 (9 blocks), 6x6 (4 blocks), 3x3 (1 block), 1x1 (1 block)

Therefore only the top-most level is going to be the same number of blocks.

walbourn:
Note that normally if you use ``fixbc4x4`` it will resize, regenerate mips, and then recompress. This is the general fix.

There's a special case that if you started wth a BC file that was less than 4x4, then it never generate mips since it doesn't add any real value.

PR #162, opened 2020-01-21
Ignore CMake generated files

Akarinnnnn:


walbourn:
In my use all the stuff ends up under ``/out`` so is already ignored..

Akarinnnnn:
> In my use all the stuff ends up under /out so is already ignored..  

Well, I mean files in `CMakeFiles`, they are not ignoed.

Akarinnnnn:
and these files
image

walbourn:
None of those files end up in my directory when using CMake. They always end up in ``./out/..``

PR #165, opened 2020-02-22
Support reading files that use a 'BC7\0' FourCC

0xC0000054:
There are files that use this FourCC in the wild, but I do not know what tool was used to create them.


walbourn:
Can you please upload an example ``DDS`` file as an attachment? I want to have a test case for my test suite.

0xC0000054:
> Can you please upload an example `DDS` file as an attachment? I want to have a test case for my test suite.

Done.

[EM2_d.zip](https://github.com/microsoft/DirectXTex/files/4242456/EM2_d.zip)

PR #179, opened 2020-05-29
Switched to using strongly typed flags

walbourn:
There are many different flags in DirectXTex, and they all use DWORD so it's easy to get the wrong flag.

This PR changes all the flags in DirectXTex's public interface to strongly-typed flags via the ``DEFINE_ENUM_FLAG_OPERATORS`` macro from winnt.h--this is what Direct3D uses to type flags.

**This is a breaking change** but it's fairly easy to fix them up as you can see in the ``texconv`` et al. tools. You can also check out the [test suite](https://github.com/walbourn/directxtextest/pull/13) for examples.

### DIFF
--- a/Texassemble/texassemble.cpp
+++ b/Texassemble/texassemble.cpp
@@ -928,9 +928,9 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
     size_t height = 0;
 
     DXGI_FORMAT format = DXGI_FORMAT_UNKNOWN;
-    DWORD dwFilter = TEX_FILTER_DEFAULT;
-    DWORD dwSRGB = 0;
-    DWORD dwFilterOpts = 0;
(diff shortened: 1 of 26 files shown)


Texassemble/texassemble.cpp
```diff
@@ -928,9 +928,9 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
     size_t height = 0;
 
     DXGI_FORMAT format = DXGI_FORMAT_UNKNOWN;
-    DWORD dwFilter = TEX_FILTER_DEFAULT;
-    DWORD dwSRGB = 0;
-    DWORD dwFilterOpts = 0;
+    TEX_FILTER_FLAGS dwFilter = TEX_FILTER_DEFAULT;
```
walbourn, line 931: This is the kind of impact I expect it to have on client code...

Issue #104, opened 2018-06-11
[Question] How to put texconv output file into their original location?


admiralmikan:
I need help with texconv command. Whenever I perform texconv.exe -m 1 -y -r *_AM.dds to delete mip maps on my .DDS files, the output file will be put on where my texconv is located. Do you know extra command that will forced texconv to put output file into their original location, thus replace their original files, too? Since I use -r command that will search for subfolders too.

Thank you very much!

![2018-06-11 08_54_52-british - copy edit](https://user-images.githubusercontent.com/17843605/41210091-a83f30a0-6d59-11e8-89f8-9c23be786f85.jpg)

walbourn:
The ``texconv`` tool outputs to either the 'current working directory' or to the path specified by the `-o-` parameter. See the [wiki](https://github.com/Microsoft/DirectXTex/wiki/Texconv).

SearcherNo81:
For anyone having trouble getting -o to work: Using a quoted output path with a trailing backslash will result in incorrectly placed files. The backslash will be read as an escape character for the quote and the output path will be ignored. -o "C:\path\\" will not work, but -o "C:\path" will.

walbourn:
Note that the tools also support 'linux' style path separators, and you can use `--` by itself to disable command-switch processing so you can start a path with a `/`. That doesn't necessarily help with the `-o` case since that's a switch, but it can be used for input file lists.

PR #184, opened 2020-06-25
Added -ip flag to save new files in place (same folder as originals)

CodeAngry:
1.) Added -ip flag to save new files in place (same folder as originals).
In case names collide, a _ is appended to the file name.
2.) Added a basic method of testing custom command lines in the debugger.

### DIFF
--- a/Texconv/texconv.cpp
+++ b/Texconv/texconv.cpp
@@ -112,6 +112,7 @@ namespace
         OPT_ROTATE_COLOR,
         OPT_PAPER_WHITE_NITS,
         OPT_BCNONMULT4FIX,
+        OPT_INPLACE,
         OPT_MAX
     };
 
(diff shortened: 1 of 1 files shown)


walbourn:
So the intent of the original design of ``texconv`` is that if you want the files to go the same logical, you can add a prefix or a post-fix with ``-px`` or ``-sx``.

CodeAngry:
It's a fully optional feature. And I prefer to convert the files in place and then run a second tool that bundles them together into another format based on the folders where they are located. But finding them in their original location is much easier and makes more sense. And did I mention?, it's opt-in.

It's also been requested before so I'm definitely not the only one to need this:
https://github.com/Microsoft/DirectXTex/issues/104

Code change is simple, easy to follow. And this is not about intent of the tool, it's about this is about handling more use cases and allowing more customization.

walbourn:
I respect your preference for workflow, but I think that [this](https://github.com/microsoft/DirectXTex/issues/200) might be a better general solution where you can reproduce a directory structure in a target folder

walbourn:
The "in-place" approach is pretty dangerous in terms of potentially losing your original files. With the new ``-r:keep`` option, you should be able to keep any directory structure you want in place in the output directory. You can always set the output directory to your source directory if you want.

PR #188, opened 2020-07-09
DirectXTex updated with DDS_FLAGS_ALLOW_LARGE_FILES

walbourn:
The DDSTextureLoader modules have checks for 'maximum required' size bounds for rejecting extremely large ``.DDS`` files that probably won't work with Direct3D at runtime. DIrectXTex's DDS codec being intended for tools didn't have such checks in case you wanted to convert some weirdly large .DDS file.

While I generally encourage using DDSTextureLoader for run-time texture loading scenarios, some people prefer to use DirectXTex. Also for viewer tools, you probably don't want to try to load unbounded size images anyhow.

This PR adds bounds checks to the DDS codec, but you can opt-out of these with a new loading flag for tools where you want to support these. I use this new flag for the command-line tools, although in practice only the x64 native versions of those tools are likely to be able to handle such large files anyhow.

### DIFF
--- a/DirectXTex/DirectXTex.h
+++ b/DirectXTex/DirectXTex.h
@@ -172,6 +172,8 @@ namespace DirectX
         DDS_FLAGS_FORCE_DX9_LEGACY      = 0x40000,
             // Force use of legacy header for DDS writer (will fail if unable to write as such)
 
+        DDS_FLAGS_ALLOW_LARGE_FILES     = 0x1000000,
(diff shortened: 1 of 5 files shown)


walbourn:
Note for the test suite, none of my tests ever exceed these values in any case. I will add the use of ``DDS_FLAGS_ALLOW_LARGE_FILES ``  to the 'fuzz tests' just so there's more chance of catching issues in the codec.

PR #196, opened 2020-09-16
Avoid static initialization order fiasco

fibann:
GUIDs in wincodec.h are defined out of line, so an initializer for a static
variable can read them before they have been initialized themselves. Taking a
reference rather than copying the value bypasses this issue.

This might turn out not to be a problem in most cases, but is still undefined
behavior. Anecdotally, removing it fixed a startup crash I got on incremental
builds (so there might be some weird interaction with the linker). So it is
probably worth fixing anyway.

### DIFF
--- a/WICTextureLoader/WICTextureLoader11.cpp
+++ b/WICTextureLoader/WICTextureLoader11.cpp
@@ -63,7 +63,7 @@ namespace
     //-------------------------------------------------------------------------------------
     struct WICTranslate
     {
-        GUID                wic;
+        const GUID&         wic;
         DXGI_FORMAT         format;
     };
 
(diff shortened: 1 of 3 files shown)


walbourn:
The GUIDs are actually in ``uuid.lib`` as data segments, so this is not a problem. That said, I'll take a look at the code-gen here to see if this is better...

walbourn:
This looks good if you also change the arrays that use these structs as ``constexpr``.

walbourn:
Updated *DirectX Tool Kit* in [this commit](https://github.com/microsoft/DirectXTK/commit/cd4175f6f9917772d8923561100261fa3895160d) and [this one](https://github.com/microsoft/DirectXTK12/commit/1a274429b424fbc559582e31a261ae3212b7c1c5)

walbourn:
Merged with additional code review in [this commit](https://github.com/microsoft/DirectXTex/commit/216e2b0388b2e954e394e9dfaebc0c0ed7688af0)

PR #197, opened 2020-09-21
Added TGA_FLAGS for TGA reader

walbourn:
Normally the TGA reader handles the case of finding an all zero alpha channel to mean "opaque" which results in setting all the alpha values to 255. This is fine and deals with some wierd older TGA files, unless you really want an all zero alpha channel.

This adds a new flag ``TGA_FLAGS_ALLOW_ALL_ZERO_ALPHA`` to let you disable this behavior. ``TGA_FLAGS_NONE`` keeps the original behavior.

Also added ``TGA_FLAGS_BGR`` so you can get back BGRX (for 24bpp TGAs) or BGRA (32bpp TGAs) instead of always getting back RGBA.

> The TGA reader has had no explicit option flags before now. I've added the flags to all the TGA functions to be consistent with DDS/WIC, although save don't use the flags yet. Also added inlines so older clients do not break.

### DIFF
--- a/DirectXTex/DirectXTex.h
+++ b/DirectXTex/DirectXTex.h
@@ -192,6 +192,14 @@ namespace DirectX
             // Enables the loader to read large dimension .dds files (i.e. greater than known hardware requirements)
     };
 
+    enum TGA_FLAGS : unsigned long
+    {
+        TGA_FLAGS_NONE                 = 0x0,
+
+        TGA_FLAGS_ALLOW_ALL_ZERO_ALPHA = 0x1,
+            // If the loaded image has an all zero alpha channel, normally we assume it should be opaque. This flag leaves it alone.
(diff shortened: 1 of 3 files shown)


walbourn:
The only 'optional' behavior of the writer at the moment is included the TGA 2.0 extension area, but this is already controlled by passing the ``metadata`` parameter (which defaults to ``nullptr``).

walbourn:
Actually, the one other TGA flag that could be useful to add would be to let 24bpp come back as BGRX, and 32bpp come back as BGRA. Currently I force both cases to RGBA.

walbourn:
Test suite updates are in [this PR](https://github.com/walbourn/directxtextest/pull/14)

scm-xbox (approved):

PR #205, opened 2020-12-18
CMake package support 

walbourn:
This PR updates the CMake support in DirectXTex to support ``find_package(directxtex CONFIG)``:

```
find_package(directxtex CONFIG REQUIRED)

target_link_libraries(foo Microsoft::DirectXTex)
```


walbourn:
Related to this issue: https://github.com/microsoft/DirectXTK/issues/247

PR #207, opened 2020-12-29
Code review for standard library header usage

walbourn:
Reviewed all the C++ standard library header usage and normalized usage for improved conformance.

* ``algorithm`` was used internally, but not in the public interface
* ``size_t`` requires ``cstddef``
* ``memcpy``, ``memcpy_s`` is officially in ``cstring``
* Use ``cassert`` instead of ``assert.h``, ``ctime`` instead of ``time.h``
* ``std::move`` is in ``utility``

> The non-standard malloc.h is still being included for ``_aligned_malloc`` / ``_aligned_free``. To be conformant, I should update it to use ``aligned_alloc`` and ``free`` when building for C++17 which is in ``cstdlib``.

Also trimmed trailing whitespace in these files.

### DIFF
--- a/DDSView/ddsview.cpp
+++ b/DDSView/ddsview.cpp
@@ -9,16 +9,16 @@
 
 #define WIN32_LEAN_AND_MEAN
 #define NOMINMAX
-#include <windows.h>
-
-#include <assert.h>
-#include <stdio.h>
-#include <dxgiformat.h>
-#include <d3d11.h>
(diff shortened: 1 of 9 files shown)


walbourn:
Same fixes for 'standalone' modules in [this commit](https://github.com/microsoft/DirectXTex/commit/18c2d9a5ca5e5c0609db1d2866408cadff8c75bc)

PR #208, opened 2021-01-06
Support for Windows Subsystem for Linux

walbourn:
By using [DirectX-Headers](https://github.com/microsoft/DirectX-Headers) version v1.0.1 or later, the library can be built for Linux using cmake.

> Command-line tool is only supported for Win32 currently


jenatali (approved):

walbourn:
Hacked up the code to use ``fstream`` on Windows and verified DDS, TGA, and HDR codecs passed the test suite.

DirectXTex/DirectXTexConvert.cpp
```diff
@@ -2998,8 +2997,8 @@ uint32_t DirectX::_GetConvertFlags(DXGI_FORMAT format) noexcept
 #endif
 
     ConvertData key = { format, 0, 0 };
-    auto in = reinterpret_cast<const ConvertData*>(bsearch_s(&key, g_ConvertTable, std::size(g_ConvertTable), sizeof(ConvertData),
-        ConvertCompare, nullptr));
+    auto in = reinterpret_cast<const ConvertData*>(bsearch(&key, g_ConvertTable, std::size(g_ConvertTable), sizeof(ConvertData),
```
walbourn, line 3000: ``bsearch_s`` like most of the ``_s`` variants in the C++17 Appendex K are not supported by GNUC's library. As it turns out, in this case it doesn't matter at all. The difference with ``bsearch_s`` is that it passes a context pointer through, which we don't use here anyhow. Just simplified it to always use ``bsearch``.

DirectXTex/DirectXTexDDS.cpp
```diff
@@ -1974,11 +2063,7 @@ HRESULT DirectX::SaveToDDSMemory(
                 if (fastpath)
                 {
                     size_t pixsize = images[index].slicePitch;
-                    if (memcpy_s(pDestination, remaining, images[index].pixels, pixsize))
```
walbourn, line 1977: GNUC's library doesn't implement Appendix K stuff like ``memcpy_s``. I already do plenty with the counts to guard against issues that would trip up ``memcpy_s`` anyhow.

These were some of the few cases where I ever checked the return value of ``memcpy_s``, so in most places just using ``memcpy`` directly without any other changed worked. Here of course, it fails even when it works. Again, there's really no failure case here so using ``mempcy`` is fine.

DirectXTex/DirectXTexImage.cpp
```diff
@@ -20,6 +20,19 @@ namespace DirectX
 
 using namespace DirectX;
 
+#ifndef WIN32
+namespace
+{
+    inline void * _aligned_malloc(size_t size, size_t alignment)
```
walbourn, line 26: The Microsoft Standard C++ Library doesn't currently implement this C+17 ``std::aligned_alloc``. That because the C++17 standard version is freed with ``free`` instead of a specific function, and on Windows these are all thin wrappers over the Win32 Heap API.

PR #210, opened 2021-01-09
texconv -swizzle option

walbourn:
Allows arbitrary swizzles for source data:

```
texconv -swizzle rgba <filename>
```

```
texconv -swizzle rrra <filename>
```

```
texconv -swizzle bgra <filename>
```

```
texconv -swizzle aaaa <filename>
```

```
texconv -swizzle zzwx <filename>
```

etc.


walbourn:
Also support [HLSL-style](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx9-graphics-reference-asm-ps-registers-modifiers-source-register-swizzling) masks where you repeat the last used component for less than 4 channels

````
texconv -swizzle rg <filename>

// same as rggg
```

````
texconv -swizzle a <filename>

// same as aaaa
```

PR #211, opened 2021-01-10
texassemble updated with -swizzle option for merge

walbourn:
Texassemble's merge was originally created to match legacy DXTex.exe behavior for "Open Alpha onto surface". This just pulled the Blue channel from the second image and placed it the Alpha channel of the first image preserving RGB.

To make this more general, there's now an optional ``-swizzle`` parameter for texassemble merge.  Lower case letters in the mask indicate the channel to use from the first image. Upper-case characters in the mask indicate channel to use from the second image.

```
texassemble merge -swizzle rrrR
```

This results in the Red channel of the first image placed into Red, Green, and Blue of the output. The Red channel of the second image image would be in the Alpha of the output.

```
texassemble merge -swizzle ggbA
```

This results in the Green channel of the first image placed into Red and Green. Blue from the first image into the Blue of the output. The Alpha channel of the second image image would be in the Alpha of the output.

```
texassemble merge -swizzle rGbA
```

This results in the Red and Blue channels of the first image merged with the Green and Alpha channels of the second image.

The default matches the original behavior: ``-swizzle rgbB``.

### DIFF
--- a/Texassemble/texassemble.cpp
+++ b/Texassemble/texassemble.cpp
@@ -96,6 +96,7 @@ namespace
         OPT_FEATURE_LEVEL,
         OPT_TONEMAP,
         OPT_GIF_BGCOLOR,
+        OPT_SWIZZLE,
         OPT_MAX
     };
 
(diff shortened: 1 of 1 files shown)


opusGlass:
Thank you for this @walbourn , I finally got around to testing it and it works great!

PR #218, opened 2021-02-18
Fix of CaptureTexture for Reserved and MultiSample resources

drinkertea:
#### Fix of reserved resources extraction.
It is unable to call `GetHeapProperties` for reserved resources. But such info required only for `READBACK` resources optimization, capture code should not fail for reserved resources in this case.
```
D3D12 ERROR: ID3D12Resource2::ID3D12Resource::GetHeapProperties: GetHeapProperties can not be called on a resource created with the CreateReservedResource API. [ RESOURCE_MANIPULATION ERROR #901: GETHEAPPROPERTIES_INVALIDRESOURCE]
```

#### Fix of multisample resources extraction.
Invalid desc for resolved resource copied from the original multisample resource with `Alignment = D3D12_DEFAULT_MSAA_RESOURCE_PLACEMENT_ALIGNMENT`, which fails creation call.

### DIFF
--- a/DirectXTex/DirectXTexD3D12.cpp
+++ b/DirectXTex/DirectXTexD3D12.cpp
@@ -141,12 +141,6 @@ namespace
             return HRESULT_E_NOT_SUPPORTED;
         }
 
-        D3D12_HEAP_PROPERTIES sourceHeapProperties;
-        D3D12_HEAP_FLAGS sourceHeapFlags;
-        HRESULT hr = pSource->GetHeapProperties(&sourceHeapProperties, &sourceHeapFlags);
-        if (FAILED(hr))
-            return hr;
-
         numberOfResources = (desc.Dimension == D3D12_RESOURCE_DIMENSION_TEXTURE3D)
                             ? 1u : desc.DepthOrArraySize;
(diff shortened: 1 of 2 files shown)


walbourn:
Thanks for the fixes.

Can you update ``DirectXTex\ScreenGrab\ScreenGrab12.cpp`` as well with this same fix so we can get both in this PR?

Also if you are interested, you can submit a PR for [DirectX Tool Kit for DX12](https://github.com/microsoft/DirectXTK12) ``ScreenGrab.cpp``. If not, I'll take care of it.

walbourn (approved):
If you can also update ScreenGrab12.cpp, that would be great. Otherwise I'll take care of it.

walbourn:
Per [Microsoft Docs: ID3D12Resource::GetHeapProperties](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12resource-getheapproperties):

> his method only works on placed and committed resources, not on reserved resources. If the resource was created as reserved, E_INVALIDARG is returned. The pages could be mapped to none, one, or more heaps.

PR #238, opened 2021-06-05
Fix CONV_FLAGS_NOALPHA flag for 16bpp format

maksqwe:
Previously condition (metadata.format == DXGI_FORMAT_B5G6R5_UNORM) is always false so
check format before replace it with DXGI_FORMAT_R8G8B8A8_UNORM


walbourn (approved):

walbourn:
Thanks. Looks like a good fix.

PR #258, opened 2021-11-20
Fix shaders not compiling with spaces in the path

AnthonyFuller:
When compiling shaders, if your file path contains spaces it will fail to compile shaders, this PR fixes that.


walbourn:
Thanks for the fix. I did this change to help meeting some new vcpkg requirements, but had a lot of problems with quotes there. I'll try this one out and see if it fixes those problems.

walbourn (approved):

walbourn:
Fixed the same problem in *DirectX Tool Kit*

https://github.com/microsoft/DirectXTK/commit/89db7aa40a57a85b5f5921ecd15526aaf6088d97

https://github.com/microsoft/DirectXTK12/commit/e33307720d1152d8778012c06972dd52c48cb89e

Looks like it works great for vcpkg too. Much appreciated!

AnthonyFuller:
No problem, glad I could help!

PR #259, opened 2021-11-23
Remove inline specifier from constexpr functions

cristeigabriel:
I quote:
"A function or static data member declared with the constexpr or consteval specifier is implicitly an inline function or variable ([dcl.inline])"
from: https://eel.is/c++draft/dcl.constexpr

Also, analyzing something alike the following:

```
#include <cstdio>
#include <random>

struct Type {
  int a;
};

constexpr auto operator+(const Type& a, const Type& b) {
  return a.a + b.a;
}

int main()
{
	auto a = Type{rand()};
  	auto b = Type{rand()};
  	auto c = a + b;
}
```

with C++ Insights, inline will be implied regardless, so there's no reason for it to be in the code base itself.

### DIFF
--- a/DirectXTex/DirectXTex.inl
+++ b/DirectXTex/DirectXTex.inl
@@ -26,16 +26,16 @@ DEFINE_ENUM_FLAG_OPERATORS(CNMAP_FLAGS);
 DEFINE_ENUM_FLAG_OPERATORS(CMSE_FLAGS);
 
 // WIC_FILTER modes match TEX_FILTER modes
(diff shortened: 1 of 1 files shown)


walbourn (approved):
Thanks. I did a cleanup pass for these last month, but I realize now that I only searched *.h and *.cpp files, and forgot to check *.inl.

Thanks!

walbourn:
One more case [here](https://github.com/microsoft/DirectXTex/commit/ab6776554bba3f7276c199dceb8f0b6816846e68).

cristeigabriel:
Sorry - didn't notice it

PR #268, opened 2022-03-02
Add support for HEIF and WEBP to command-line tools

walbourn:
HEIF/HEIC and WEBP are not supported by 'native' WIC codecs, but are supported by official Microsoft installable codecs from the Microsoft Store. This updates the command-line tools to support these.

> WEBP is only supported for reading. HEIF/HEIC is supported for read and write.

### DIFF
--- a/DirectXTex/DirectXTexUtil.cpp
+++ b/DirectXTex/DirectXTexUtil.cpp
@@ -270,6 +270,10 @@ REFGUID DirectX::GetWICCodec(WICCodecs codec) noexcept
     case WIC_CODEC_ICO:
         return GUID_ContainerFormatIco;
 
+    case WIC_CODEC_HEIF:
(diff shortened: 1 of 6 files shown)


DirectXTex/DirectXTexUtil.cpp
```diff
@@ -270,6 +270,10 @@ REFGUID DirectX::GetWICCodec(WICCodecs codec) noexcept
     case WIC_CODEC_ICO:
         return GUID_ContainerFormatIco;
 
+    case WIC_CODEC_HEIF:
+        // This requires installing https://aka.ms/heif
+        return GUID_ContainerFormatHeif;
```
walbourn, line 275: This symbol exists in the Windows 10 SDK (19041) which is the oldest SDK I support.

scm-xbox (approved):

PR #275, opened 2022-05-01
CMake support added for building with MinGW

walbourn:
Since DirectXTex already supports building using GNUC for WSL, added CMakePresets for using MinGW as well:

```
x64-Debug-MinGW
x64-Release-MinGW
x86-Debug-MinGW
x86-Release-MinGW
```

Makes use of VCPKG for DirectXMath and DirectX-Headers since the MinGW has minimal versions of these libraries. Requires the ``VCPKG_ROOT`` envvar set.

This is being done to improve compiler conformance for the code.

> Also normalized preprocessor define use to ``_WIN32`` and not ``WIN32``.


CMakeLists.txt
```diff
@@ -154,7 +154,7 @@ if(ENABLE_OPENEXR_SUPPORT)
   target_include_directories(${PROJECT_NAME} PRIVATE ${OPENEXR_INCLUDE_DIRS}/OpenEXR)
 endif()
 
-if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16")
+if ((${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16") AND (NOT MINGW))
```
walbourn, line 157: Hit some kind of ICE in MinGW with PCH, so disabled it.

DDSView/ddsview.cpp
```diff
@@ -452,8 +452,8 @@ HRESULT InitDevice(const TexMetadata& mdata)
     vp.TopLeftX = vp.TopLeftY = 0.f;
     vp.Width = static_cast<float>(width);
     vp.Height = static_cast<float>(height);
-    vp.MinDepth = D3D11_MIN_DEPTH;
```
walbourn, line 455: MinGW's d3d11 headers are pretty quirky. They are missing the ``D3D11_MIN_DEPTH`` and ``D3D11_MAX_DEPTH`` symbols.

DirectXTex/DirectXTexMipmaps.cpp
```diff
@@ -543,7 +543,7 @@ HRESULT DirectX::Internal::ResizeSeparateColorAndAlpha(
 
         if (SUCCEEDED(hr))
         {
-            WICInProcPointer colorWithAlphaData = nullptr;
```
walbourn, line 546: Another type missing from the MinGW headers, but this one is just a typedef for ``BYTE *`` which is a lot clearer anyhow.

DirectXTex/DirectXTexP.h
```diff
@@ -181,6 +185,10 @@ using WICPixelFormatGUID = GUID;
 
 #define XBOX_DXGI_FORMAT_R4G4_UNORM DXGI_FORMAT(190)
 
+#if defined(__MINGW32__) && !defined(E_BOUNDS)
+#define E_BOUNDS static_cast<HRESULT>(0x8000000BL)
```
walbourn, line 189: Another missing symbol in MinGW headers.

PR #280, opened 2022-07-25
Enable guard:ehcont compiler flag for MSVC 64 bit architectures

SergioRZMasson:
Enabled guard:ehcont so the next nuget packages releases bring binaries compiled with that flag. Enabling this flag should not have any impact if the library consumer does not enable it. However, without it is not possible for consumers that want to use the flag to link against DirectXTex.

For more information see: 
https://docs.microsoft.com/en-us/cpp/build/reference/guard-enable-eh-continuation-metadata?view=msvc-170


walbourn:
Note: This requires VS 2019 16.7 or later, although that's not a huge issue at this point.

CMakeLists.txt
```diff
@@ -173,6 +173,12 @@ if(MSVC)
     string(REPLACE "/GR " "/GR- " CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
     string(REPLACE "/GR " "/GR- " CMAKE_CXX_FLAGS_DEBUG ${CMAKE_CXX_FLAGS_DEBUG})
     string(REPLACE "/GR " "/GR- " CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE})
+
+    # Compiling with guard:ehcont allow library consumers to enable guard:ehcont if they want. (Requires VS 2019 x64)
+    if((MSVC_VERSION GREATER 1920) AND (CMAKE_SIZEOF_VOID_P EQUAL 8))
```
walbourn, line 178: Make this ``1928`` since this requires VS 2019 16.7 or later.

walbourn: Also, what happens here when building for ARM64? Also the target name here is wrong, so I'm guessing this was a cut & paste.

This should be moved down to the ``if(MSVC)`` section with other such conditionals.

```
if(MSVC)
    foreach(t IN LISTS TOOL_EXES ITEMS ${PROJECT_NAME})
      target_compile_options(${t} PRIVATE /fp:fast "$<$<NOT:$<CONFIG:DEBUG>>:/guard:cf>")
      target_link_options(${t} PRIVATE /DYNAMICBASE /NXCOMPAT)
    endforeach()

    if((${CMAKE_SIZEOF_VOID_P} EQUAL 4) AND (NOT ${DIRECTX_ARCH} MATCHES "^arm"))
        foreach(t IN LISTS TOOL_EXES ITEMS ${PROJECT_NAME})
          target_link_options(${t} PRIVATE /SAFESEH)
        endforeach()
    endif()

    if((MSVC_VERSION GREATER_EQUAL 1928) AND (${CMAKE_SIZEOF_VOID_P} EQUAL 8) AND (NOT ${DIRECTX_ARCH} MATCHES "^arm"))
        foreach(t IN LISTS TOOL_EXES ITEMS ${PROJECT_NAME})
          target_link_options(${t} PRIVATE "$<$<CONFIG:RELEASE>:/guard:ehcont>")
        endforeach()
    endif()
endif()
```

DirectXTex/DirectXTex_Desktop_2019.vcxproj
```diff
@@ -170,6 +170,8 @@
       <ExternalWarningLevel>Level4</ExternalWarningLevel>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
       <SupportJustMyCode>false</SupportJustMyCode>
+      <GuardEHContMetadata>
```
walbourn, line 173: Use notepad to clean up these blank elements.

DirectXTex/DirectXTex_Desktop_2022.vcxproj
```diff
@@ -170,6 +170,8 @@
       <ExternalWarningLevel>Level4</ExternalWarningLevel>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
       <SupportJustMyCode>false</SupportJustMyCode>
+      <GuardEHContMetadata>
```
walbourn, line 173: Is this already defaulted correctly in VS 2022?

walbourn (changes requested):
Once we get this all wrapped up, I'll make sure to update DirectXTK, DirectXTK12, DirectXMesh, UVAtlas, etc. to match.

CMakeLists.txt
SergioRZMasson: Documentation says that /guard:ehcont is "supported for 64-bit processes on a 64-bit OS" so I'm assuming ARM64 is also supported.

CMakeLists.txt
```diff
@@ -297,6 +297,12 @@ if(MSVC)
           target_link_options(${t} PRIVATE /SAFESEH)
         endforeach()
     endif()
+
+    # Compiling with guard:ehcont allow library consumers to enable guard:ehcont if they want. (Requires VS 2019 16.7 or later)
+    if((MSVC_VERSION GREATER 1928) AND (CMAKE_SIZEOF_VOID_P EQUAL 8))
```
walbourn, line 302: ``GREATER_EQUAL``

DirectXTex/DirectXTex_Desktop_2022.vcxproj
SergioRZMasson: Didn't found any information about the it been default on VS 2022. I don't think it would be since enabling it requires your dependencies to also have it enabled which would break projects that consume precompiled binaries.

walbourn: It appears to be "undefined" by default, neither true or false. Explicit adding it to Release|x64 and Release|ARM64 (and maybe Profile|x64/arm64) seems reasonable.

walbourn (changes requested):
One more fix for CMake and then we are good.

CMakeLists.txt
```diff
@@ -297,6 +297,12 @@ if(MSVC)
           target_link_options(${t} PRIVATE /SAFESEH)
         endforeach()
     endif()
+
+    # Compiling with guard:ehcont allow library consumers to enable guard:ehcont if they want. (Requires VS 2019 16.7 or later)
+    if((MSVC_VERSION GREATER_EQUAL 1928) AND (CMAKE_SIZEOF_VOID_P EQUAL 8))
+      target_compile_options(${PROJECT_NAME} PRIVATE "$<$<CONFIG:RELEASE>:/guard:ehcont>")
```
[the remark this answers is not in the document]
walbourn: I'll do this fix and submit. Thanks!

walbourn:
https://github.com/microsoft/DirectXMesh/commit/cc193564bb047cac5adcfc12a38bfd2c66043e10

https://github.com/microsoft/DirectXTK/commit/03918f87084a4b61c9fed9f73e90aeae6b998e83

https://github.com/microsoft/DirectXTK12/commit/f34a45dca934a6205f327340d2aa077f4dd0524e

https://github.com/microsoft/UVAtlas/commit/946d0c9b5c392e8b55c1f2186ac063a12a6c2423

walbourn:
This builds fine with VS 2019 MSVC, VS 2022 MSVC, and VS 2022 Clang v13. It fails with VS 2019 Clang v12 via CMake...

SergioRZMasson:
Thanks for helping with that @walbourn ! What would be the criteria for publishing new nuget packages with these flags? Are you guys the ones publishing it?

walbourn:
I'm gearing up for a release cycle for my libs this week as I have a few other fixes accumulated as well, so it's good timing.

I build the static library for the NuGet packages currently using the 14.28.29910 toolset so it's binary compatible with VS 2019 16.9 or later in keeping with the current VS 2019 lifecycle schedule. Therefore, they will be built with ``/guard:ehcont`` with these changes.

PR #290, opened 2022-11-06
Texassemble updates for cubemap generation from a image

matyalatte:
* Added commands to create cubemap from a image
    `cube-from-hc` for h-cross layout
    `cube-from-vc` for v-cross layout
    `cube-from-vc-fnz` for v-cross layout flipping the -Z face
    `cube-from-ht` for h-tee layout
    `cube-from-hs` for h-strip layout
    `cube-from-vs` for v-strip layout
* Changed help message for array-strip (from "creates ..." to "create ...")

### DIFF
--- a/Texassemble/texassemble.cpp
+++ b/Texassemble/texassemble.cpp
@@ -79,6 +79,12 @@ namespace
         CMD_MERGE,
         CMD_GIF,
         CMD_ARRAY_STRIP,
+        CMD_CUBE_FROM_HC,
+        CMD_CUBE_FROM_VC,
+        CMD_CUBE_FROM_VC_FNZ,
+        CMD_CUBE_FROM_HT,
+        CMD_CUBE_FROM_HS,
+        CMD_CUBE_FROM_VS,
         CMD_MAX
     };
 
@@ -130,19 +136,25 @@ namespace
 
     const SValue g_pCommands[] =
     {
-        { L"cube",          CMD_CUBE },
-        { L"volume",        CMD_VOLUME },
-        { L"array",         CMD_ARRAY },
-        { L"cubearray",     CMD_CUBEARRAY },
-        { L"h-cross",       CMD_H_CROSS },
-        { L"v-cross",       CMD_V_CROSS },
-        { L"v-cross-fnz",   CMD_V_CROSS_FNZ },
-        { L"h-tee",         CMD_H_TEE },
-        { L"h-strip",       CMD_H_STRIP },
-        { L"v-strip",       CMD_V_STRIP },
-        { L"merge",         CMD_MERGE },
-        { L"gif",           CMD_GIF },
-        { L"array-strip",   CMD_ARRAY_STRIP },
+        { L"cube",              CMD_CUBE },
+        { L"volume",            CMD_VOLUME },
+        { L"array",             CMD_ARRAY },
+        { L"cubearray",         CMD_CUBEARRAY },
+        { L"h-cross",           CMD_H_CROSS },
+        { L"v-cross",           CMD_V_CROSS },
+        { L"v-cross-fnz",       CMD_V_CROSS_FNZ },
+        { L"h-tee",             CMD_H_TEE },
+        { L"h-strip",           CMD_H_STRIP },
+        { L"v-strip",           CMD_V_STRIP },
+        { L"merge",             CMD_MERGE },
+        { L"gif",               CMD_GIF },
+        { L"array-strip",       CMD_ARRAY_STRIP },
+        { L"cube-from-hc",      CMD_CUBE_FROM_HC },
+        { L"cube-from-vc",      CMD_CUBE_FROM_VC },
+        { L"cube-from-vc-fnz",  CMD_CUBE_FROM_VC_FNZ },
+        { L"cube-from-ht",      CMD_CUBE_FROM_HT },
+        { L"cube-from-hs",      CMD_CUBE_FROM_HS },
+        { L"cube-from-vs",      CMD_CUBE_FROM_VS },
         { nullptr,          0 }
     };
 
@@ -741,9 +753,15 @@ namespace
             L"   v-cross-fnz         create a cross image flipping the -Z face\n"
             L"   h-tee               create a 'T' image from a cubemap\n"
             L"   h-strip or v-strip  create a strip image from a cubemap\n"
-            L"   array-strip         creates a strip image from a 1D/2D array\n"
+            L"   array-strip         create a strip image from a 1D/2D array\n"
             L"   merge               create texture from rgb image and alpha image\n"
             L"   gif                 create array from animated gif\n"
+            L"   cube-from-hc        create cubemap from a h-cross image\n"
+            L"   cube-from-vc        create cubemap from a v-cross image\n"
+            L"   cube-from-vc-fnz    create cubemap from a v-cross image flipping the -Z face\n"
+            L"   cube-from-ht        create cubemap from a h-tee image\n"
+            L"   cube-from-hs        create cubemap from a h-strip image\n"
+            L"   cube-from-vs        create cubemap from a v-strip image\n"
             L"\n"
             L"   -r                  wildcard filename search is recursive\n"
             L"   -flist <filename>   use text file with a list of input files (one per line)\n"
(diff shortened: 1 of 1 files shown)


matyalatte:

Texassemble/texassemble.cpp
```diff
@@ -2240,6 +2282,205 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
...
+            auto src = loadedImages.cbegin();
+            auto img = (*src)->GetImage(0, 0, 0);
+            size_t twidth = 0;
+            size_t theight = 0;
+
+            switch (dwCommand)
+            {
+            case CMD_CUBE_FROM_HC:
+            case CMD_CUBE_FROM_HT:
+                twidth = width / 4;
```
walbourn, line 2302: A good safety check in each of these would be:

```
if (((width % 4) != 0) || ((height % 3) != 0))
{
    wprintf(L"WARNING: cube-from-hc expects 4:3 aspect ratio\n");
}
```

3:4, 6:1, 1:6, etc.

Texassemble/texassemble.cpp
```diff
@@ -1007,10 +1025,16 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
...
+    case CMD_CUBE_FROM_VC:
+    case CMD_CUBE_FROM_VC_FNZ:
+    case CMD_CUBE_FROM_HT:
+    case CMD_CUBE_FROM_HS:
+    case CMD_CUBE_FROM_VS:
         break;
 
     default:
-        wprintf(L"Must use one of: cube, volume, array, cubearray,\n   h-cross, v-cross, v-cross-fnz, h-tee, h-strip, v-strip,\n   array-strip, merge, gif\n\n");
+        wprintf(L"Must use one of: cube, volume, array, cubearray,\n   h-cross, v-cross, v-cross-fnz, h-tee, h-strip, v-strip,\n   array-strip, merge, gif,\n   cube-from-hc, cube-from-vc, cube-from-vc-fnz,\n   cube-from-ht, cube-from-hs, cube-from-vs\n\n");
```
walbourn, line 1037: This is getting unwieldy to maintain. Replace it with:

```
        wprintf(L"Must use one of:");
        for (size_t j = 0; j < std::size(g_pCommands); ++j)
        {
            if (!g_pCommands[j].name)
            {
                wprintf(L"\n");
                break;
            }

            wprintf(L" %ls", g_pCommands[j].name);
            if ((j % 4) == 3)
                wprintf(L"\n\t");
        }
        return 1;
```

Texassemble/texassemble.cpp
```diff
@@ -2240,6 +2282,205 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
...
+            }
+
+            ScratchImage result;
+            hr = result.InitializeCube(format, twidth, theight, 1, 1);
+            if (FAILED(hr))
+            {
+                wprintf(L"FAILED setting up result image (%08X%ls)\n", static_cast<unsigned int>(hr), GetErrorDesc(hr));
+                return 1;
+            }
+            memset(result.GetPixels(), 0, result.GetPixelsSize());
```
walbourn, line 2338: You shouldn't need to set it to 0 since you are filling the entirely of each face.

walbourn: It turns out ``InitializeFromCube`` is already zeroing it anyhow I don't need it at line 1965 either :)

walbourn (changes requested):

Texassemble/texassemble.cpp
matyalatte: You mean I should remove all memset lines from this file?
There are 2 memset lines for `Initialize2D`.

walbourn: Anywhere ``memset`` is used right after a ``ScratchImage::Initialize*`` is redundant.

Texassemble/texassemble.cpp
matyalatte: I noticed there is `PrintList`.

```
wprintf(L"Must use one of: ");
PrintList(17, g_pCommands);
return 1;
```
would be better?

Texassemble/texassemble.cpp
matyalatte: I'll define ratio instead of size with switch statement.
Is that OK?

```c++
size_t ratio_w = 1;
size_t ratio_h = 1;

switch (dwCommand)
{
case CMD_CUBE_FROM_HC:
case CMD_CUBE_FROM_HT:
    // 4:3 aspect ratio
    ratio_w = 4;
    ratio_h = 3;
    break;

...

}

size_t twidth = width / ratio_w;
size_t theight = height / ratio_h;

if (((width % ratio_w) != 0) || ((height % ratio_h) != 0) || (twidth != theight))
{
    wprintf(L"WARNING: %ls expects %d:%d aspect ratio\n", dwCommand, ratio_w, ratio_h);
}
```

Texassemble/texassemble.cpp
walbourn: Right. ``PrintList(4, g_pCommands);`` should work (doesn't need quite as much spacing at the front.

Texassemble/texassemble.cpp
walbourn: Looks like that should work. Just need to make sure the warning isn't emitted for other kinds of commands.

matyalatte: You're right. But it's already conditioned by another switch statement.
So, I'll use my code as it is.

matyalatte:
Thanks for the review.
I fixed the requested parts at https://github.com/microsoft/DirectXTex/pull/290/commits/15c9deec934989d053bfa6bcb6eb5c457358eda5.

Texassemble/texassemble.cpp
```diff
@@ -1244,10 +1269,16 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
...
+                case CMD_CUBE_FROM_VC:
+                case CMD_CUBE_FROM_VC_FNZ:
+                case CMD_CUBE_FROM_HT:
+                case CMD_CUBE_FROM_HS:
+                case CMD_CUBE_FROM_VS:
                     break;
 
                 default:
-                    wprintf(L"-stripmips only applies to cube, volume, array, cubearray, or merge commands\n");
+                    wprintf(L"-stripmips only applies to cube, volume, array, cubearray, merge, or cube-from-* commands\n");
```
walbourn, line 1281: I don't think ``-stripmips`` does anything for your new functions.

Texassemble/texassemble.cpp
```diff
@@ -2240,6 +2279,210 @@ int __cdecl wmain(_In_ int argc, _In_z_count_(argc) wchar_t* argv[])
...
+                break;
+
+            default:
+                break;
+            }
+
+            size_t twidth = width / ratio_w;
+            size_t theight = height / ratio_h;
+
+            if (((width % ratio_w) != 0) || ((height % ratio_h) != 0) || (twidth != theight))
```
walbourn, line 2324: I don't think the result is required to be square.

walbourn (approved):
If you can fix the two new issues, that would be awesome. Otherwise, I can accept this and do a few edits myself.

Texassemble/texassemble.cpp
matyalatte: Right. so I'll disable another error at line 1501 as well.
like this
```c++
else if ((info.mipLevels > 1) && ((dwOptions & (1 << OPT_STRIP_MIPS)) == 0))
{
    switch (dwCommand)
    {
    case CMD_CUBE_FROM_HC:
    case CMD_CUBE_FROM_VC:
    case CMD_CUBE_FROM_VC_FNZ:
    case CMD_CUBE_FROM_HT:
    case CMD_CUBE_FROM_HS:
    case CMD_CUBE_FROM_VS:
        break;
    default:
        wprintf(L"\nERROR: Can't assemble using input mips. To ignore mips, try again with -stripmips\n");
        return 1;
        break;
   }
}
```

Texassemble/texassemble.cpp
matyalatte: No problem removing the last condition because it's just a warning.
But it's not expected ratio when the result is not square, I think.

For example, it'll allow 40:3 aspect ratio for h-cross layout.

matyalatte:
Sorry for force-pushing but I think I fixed the new issues at https://github.com/microsoft/DirectXTex/pull/290/commits/8482fbfafc5d38c8f5fe27b1f9c596688424c0fd.

walbourn:
Updated the test scripts for these cases:

https://github.com/walbourn/directxtextest/commit/2c0cd15f9a244de3ec8f40d42e8a23374d6cdc86

PR #293, opened 2022-11-25
Add MSVC Spectre mitigation  

SergioRZMasson:
Added the /Qspectre MSVC compiler option when supported (Visual Studio 2017 version 15.5.5 and all later versions). This should mitigate certain Spectre variant 1 security vulnerabilities.

More information:
https://learn.microsoft.com/en-us/cpp/build/reference/qspectre?view=msvc-170


SergioRZMasson:
@walbourn , is there anything we need to do to move this forward?

walbourn:
My general concern is that Spectre-mitgated libraries are *not* installed by default with the C++ Workload in Visual Studio, so these changes as they are will break for many people.

I think the better solution here is to do the following:

(A) For all users of the vcxproj scenario, they can easily add it externally via ``/p:SpectreMitigation=Spectre`` (by design). I'll make sure to add a note about this to my readmes/docs. I can also add a flavor of the libs to the NuGet to include this mitigation.

(B) For the CMakeLists.txt, keep the new option but have it guarded by a build option.

DirectXTex/DirectXTex_Windows10_2022.vcxproj
```diff
@@ -107,41 +107,49 @@
     <ConfigurationType>StaticLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
     <PlatformToolset>v143</PlatformToolset>
+    <SpectreMitigation>Spectre</SpectreMitigation>
```
walbourn, line 110: Note that UWP does NOT support Spectre libs.

SergioRZMasson:
BTW, we consume DirectXTex from the nuget packages. So my goal is to have the binaries in the package been built with /Qspectre flag.

walbourn:
> BTW, we consume DirectXTex from the nuget packages. So my goal is to have the binaries in the package been built with /Qspectre flag.

Understood. I'm adding a second library to the NuGet package to support this. See https://github.com/microsoft/DirectXTex/issues/294

walbourn:
CMake updates committed

https://github.com/microsoft/DirectXTex/commit/d91675bde6db83a577ffbf6f657815a333864b86

PR #296, opened 2022-12-14
CMake updated to use /Zc:inline and /Zc:lambda

walbourn:
The ``/Zc:inline`` switch enforces C++11 rules on inline visibility which reduces redundant comdats in each OBJ file (i.e. all the DirectXMath inline functions you *didn't* call in that translation unit). This switch is on by default with MSBuild, but is *not* on by default for command-line builds. This update adds this switch to CMake builds as well.

This reduces the size of the Release obj/lib files in half.

There's also a ``/Zc:lambda`` conformance switch for VS 2019 16.8 or later which I'm enabling for coverage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -374,6 +374,12 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
         endforeach()
     endif()
 
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 19.28)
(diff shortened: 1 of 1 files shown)


walbourn:
Related updates for test suite: https://github.com/walbourn/directxtextest/commit/5f6258d09fc60b765999fdb4e76b34bcbe73a694  https://github.com/walbourn/directxtextest/commit/5f6258d09fc60b765999fdb4e76b34bcbe73a694

scm-xbox (approved):

PR #303, opened 2023-01-13
Fix #350

### LINKED ISSUE #350: FlipRotate not supported for the Linux platform
The bulk of the image operations that use WIC also have non-WIC paths which are supported for Linux & Windows Subsystem for Linux. The one function that currently always requires WIC is ``FlipRotate``.

This could be implemented directly in the library as well.

UMU618:
Change RenderTargetView format DXGI_FORMAT_R8G8B8A8_UNORM_SRGB to DXGI_FORMAT_R8G8B8A8_UNORM.

### DIFF
--- a/DDSView/ddsview.cpp
+++ b/DDSView/ddsview.cpp
@@ -416,7 +416,7 @@ HRESULT InitDevice(const TexMetadata& mdata)
 
     D3D11_RENDER_TARGET_VIEW_DESC vd = {};
     vd.ViewDimension = D3D11_RTV_DIMENSION_TEXTURE2D;
-    vd.Format = DXGI_FORMAT_R8G8B8A8_UNORM_SRGB;
+    vd.Format = sd.BufferDesc.Format;   // DXGI_FORMAT_R8G8B8A8_UNORM_SRGB;
     hr = g_pd3dDevice->CreateRenderTargetView(pBackBuffer, &vd, &g_pRenderTargetView);
     pBackBuffer->Release();
     if (FAILED(hr))


walbourn:
For gamma-correct rendering, the RTV should use `DXGI_FORMAT_R8G8B8A8_UNORM_SRGB`.

The reason your texture is appearing 'too bright' is because it's not explicitly using an SRGB variant. Legacy DDS files don't indicate if they are in sRGB or linear colorspace, so assumptions are made. See the [wiki](https://github.com/microsoft/DirectXTex/wiki/DDS-I-O-Functions#related-flags) for all the various flags including ones that force SRGB formats.

UMU618:
> DDS files

My DDS files is captured from desktop by Windows Graphic Capture, and save with DDS_FLAGS_NONE flag. So I should use DDS_FLAGS_FORCE_RGB?

UMU618:
PS: I open my DDS file with Paint.NET, it's correct!

UMU618:
With NVIDIA Texture Tools Exporter 2021.2.0, also correct. Only ddsview shows 'too bright'.

walbourn:
Can you attach an example DDS file to this thread? Thanks!

UMU618:
@walbourn 
[141262599133000.zip](https://github.com/microsoft/DirectXTex/files/10416899/141262599133000.zip)

UMU618:
My code to save DDS: https://github.com/UMU618/Win32CaptureSample/blob/0c13fbd8fcdc52cb711de3bea8f2d3c9d7ecc9e0/Win32CaptureSample/SimpleCapture.cpp#L183

walbourn:
Only "DX10" header extensions can capture the ``DXGI_FORMAT_*_SRGB`` formats directly in the DDS file, but there's many reasons to want to use the legacy format when supported. Therefore, you often have to guess what the color-space is for RGBA32 formats such as this one. Typically, it's in sRGB, but not always. Most of these tools are assuming sRGB as does Visual Studio's image viewer which uses DirectXTex.

That said, **DDSView** is intended as a sample tool so I want to be very transparent about assumptions. I added a ``-forcesrgb`` switch in [this PR](https://github.com/microsoft/DirectXTex/pull/305).

PR #307, opened 2023-01-25
ConvertToSinglePlane memory overrwrite bug fix

walbourn:
**ConvertToSinglePlane** did not validate that the input images meet the height alignment requirements for the source format. This can result in a memory overwrite if untrusted content is passed from the DDS loader directly to this function.

This PR also includes additional hardening for the DDS loader, ComputePitch, and DDSTextureLoader modules to validate width & height requirements for planar video textures.


shawnhar (approved):

Flugan:
ConvertToSInglePlane bug

PR #309, opened 2023-01-30
Clean up GPU-based validation warnings

walbourn:
When **SetEnableGPUBasedValidation** is enabled on Windows, the initial state for a resource will result in a harmless but noisy warning.

```
D3D12 WARNING: ID3D12Device::CreateCommittedResource: Ignoring InitialState D3D12_RESOURCE_STATE_COPY_DEST. Buffers are effectively created in state D3D12_RESOURCE_STATE_COMMON. [ STATE_CREATION WARNING #1328: CREATERESOURCE_STATE_IGNORED]
```

The fix is to use the ``COMMON`` state initially and allow it to promote automatically.

### DIFF
--- a/DDSTextureLoader/DDSTextureLoader12.cpp
+++ b/DDSTextureLoader/DDSTextureLoader12.cpp
@@ -1283,7 +1283,7 @@ namespace
             &defaultHeapProperties,
             D3D12_HEAP_FLAG_NONE,
             &desc,
-            D3D12_RESOURCE_STATE_COPY_DEST,
+            D3D12_RESOURCE_STATE_COMMON,
             nullptr,
             IID_ID3D12Resource, reinterpret_cast<void**>(texture));
         if (SUCCEEDED(hr))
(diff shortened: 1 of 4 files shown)


walbourn:
Note that on Xbox, common state promotion is opt-in, so the *DirectX Tool Kit* for DX12 version of the loaders needs to continue to use COPY_DEST instead of COMMON.

https://github.com/microsoft/DirectXTK12/pull/149

DDSTextureLoader/DDSTextureLoader12.cpp
```diff
@@ -1283,7 +1283,7 @@ namespace
             &defaultHeapProperties,
             D3D12_HEAP_FLAG_NONE,
             &desc,
-            D3D12_RESOURCE_STATE_COPY_DEST,
```
walbourn, line 1286: The "standalone" copies of DDSTextureLoader, WICTextureLoader don't support Xbox.

walbourn:
VCPKG is having infrastructure issues today due to a git change that impacted GitHub. The code isn't a problem.

scm-xbox (approved):

PR #320, opened 2023-02-12
CMake updated to support building for Xbox

walbourn:
This adds CMake support for building for Xbox Series X|S or Xbox One via the Microsoft GDK with Xbox Extensions gaming command prompts.

Also added some conformance switches for VS 2022 17.5 and when building with WPO/LTCG via CMake.

> CompileShaders was tweaked to use the native version of FXC for the host platform.


CMakePresets.json
```diff
@@ -109,6 +109,22 @@
       },
       "hidden": true
     },
+    {
+      "name": "Scarlett",
```
walbourn, line 113: When I merge this change to the xbox branch, I'll need to rename the existing "XboxOne" and "Scarlett" to something else since those are PC builds with tooling support.

walbourn: Fixing this in [321](https://github.com/microsoft/DirectXTex/pull/321).

scm-xbox (approved):

PR #333, opened 2023-03-26
tex tool updates for POSIX

walbourn:
The **texconv**, **texassemble**, and **texdiag** tools are all Windows command-list tools, and so don't follow POSIX utility conventions. This PR, however, adds support for a few specific behaviors:

```
texconv --version
```

```
texconv --help
```

All paths read from the command-line or from -flist files are normalized to the Windows path separator, so the tool now accepts UNIX-style paths or Windows-style paths. In order to allow the input filenames to start with a `/`, you can use `--` as an escape to indicate the 'end of options' which will allow filenames to start with either `-` or `/`.

```
texconv -- /folder/path/to/name/fname.dds
```


walbourn:
https://github.com/microsoft/DirectXMesh/pull/91

scm-xbox (approved):

PR #371, opened 2023-07-06
Add support for reading Unreal Tournament 2004 DDS files

0xC0000054:
These files have an incorrectly populated DDS_PIXELFORMAT structure, all fields except the dwFourCC field are set to zero.

The following zip file contains some sample files for your test suite:  [UT2004 DDS samples.zip](https://github.com/microsoft/DirectXTex/files/11962909/UT2004.DDS.samples.zip)
The files are originally from [this](https://forums.getpaint.net/topic/22473-some-dds-files-fail-to-load-examples-included/) thread on the Paint.NET forum.

### DIFF
--- a/DirectXTex/DirectXTexDDS.cpp
+++ b/DirectXTex/DirectXTexDDS.cpp
@@ -191,10 +191,9 @@ namespace
 
         constexpr size_t MAP_SIZE = sizeof(g_LegacyDDSMap) / sizeof(LegacyDDS);
         size_t index = 0;
-        // Unreal Tournament 2004 DDS files use a incorrectly populated DDS_PIXELFORMAT
-        // structure that has all fields except the dwFourCC field set to zero.
         if (ddpf.size == 0 && ddpf.flags == 0 && ddpf.fourCC != 0)
         {
+            // Handle some DDS files where the DDPF_PIXELFORMAT is mostly zero
             for (index = 0; index < MAP_SIZE; ++index)
             {
                 const LegacyDDS* entry = &g_LegacyDDSMap[index];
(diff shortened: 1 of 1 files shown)


walbourn:
Thank you for your contribution. I'm totally open to having the DirectXTex library handle as many 'legacy' scenarios' as possible, although in some case I want to guard accepting the stranger stuff with an explicit flag to avoid introducing potential security bugs.

In this specific case, it looks like the risk here isn't much.

DirectXTex/DirectXTexDDS.cpp
```diff
@@ -191,74 +191,92 @@ namespace
 
         constexpr size_t MAP_SIZE = sizeof(g_LegacyDDSMap) / sizeof(LegacyDDS);
         size_t index = 0;
-        for (index = 0; index < MAP_SIZE; ++index)
+        // Unreal Tournament 2004 DDS files use a incorrectly populated DDS_PIXELFORMAT
```
walbourn, line 194: I think removing the specific reference to UT2004 from here and making it the following is clearer for the long-term since it may apply to other DDS files as well.

```
if (ddpf.size == 0 && ddpf.flags == 0 && ddpf.fourCC != 0)
{
    // Handle some DDS files where the DDPF_PIXELFORMAT is mostly zero
```

DirectXTex/DirectXTexDDS.cpp
```diff
@@ -321,7 +361,7 @@ namespace
 
         // Verify header to validate DDS file
         if (pHeader->size != sizeof(DDS_HEADER)
-            || pHeader->ddspf.size != sizeof(DDS_PIXELFORMAT))
+            || pHeader->ddspf.size != sizeof(DDS_PIXELFORMAT) && !IsUT2004PixelFormat(pHeader->ddspf))
```
walbourn, line 364: For this check rather than introduce the extra helper functions, I suggest just going with:

```
// Verify header to validate DDS file
if (pHeader->size != sizeof(DDS_HEADER))
{
    return E_FAIL;
}

if (pHeader->ddspf.size != 0 && pHeader->ddspf.size != sizeof(DDS_PIXELFORMAT))
{
    return E_FAIL;
}
```

walbourn (changes requested):

walbourn (approved):

walbourn:
Updated [ddsdump.cpp](https://[raw.githubusercontent.com/wiki/Microsoft/DirectXTex/ddsdump.cpp](https://raw.githubusercontent.com/wiki/Microsoft/DirectXTex/ddsdump.cpp) utility to handle this sceanrio

walbourn:
Also updated the [test suite](https://github.com/walbourn/directxtextest/commit/97b72027c082701bf9d6e0a138de0029cd1eca6d)

PR #375, opened 2023-07-09
Add CompressEx and ConvertEx methods that report progress

### LINKED ISSUE #374: Progress reporting for compression and format conversion
I implemented this in the DirectXTex fork that my Paint.NET DDS plugin uses, and I was wondering if there is any interest in adding it to the main DirectXTex distribution.
My implementation is a bit of a hack, so I am not sure how much of it could be reused.

The progress callback is defined as `bool ProgressCallback(size_t done, size_t total)`.
If the callback returns `false`, DirectXTex returns `HRESULT_FROM_WIN32(ERROR_CANCELLED)`.

I am not sure of the proposed `CompressEx` and `ConvertEx` methods should add the progress callback as a parameter, or use an options structure to simplify future expansion.
The existing `Compress` and `Convert` would call the Ex methods with the progress callback set to `nullptr`.

This should probably also be added to the `Decompress` method for consistency, but my current implementation does not have that.

0xC0000054:
This allows callers to display the progress in their user interface and/or cancel the current operation.

For single images, the progress is reported as the rows are processed.
For multiple images, the progress is reported after each image is processed.

Closes #374


DirectXTex/BCDirectCompute.cpp
```diff
@@ -371,7 +371,7 @@ HRESULT GPUCompressBC::Prepare(size_t width, size_t height, uint32_t flags, DXGI
 
 
 //-------------------------------------------------------------------------------------
-HRESULT GPUCompressBC::Compress(const Image& srcImage, const Image& destImage)
+HRESULT GPUCompressBC::Compress(const Image& srcImage, const Image& destImage, ProgressProc progressProc)
```
walbourn, line 374: Have you seen GPU compression to be slow enough to need a callback?

DirectXTex/DirectXTex.h
```diff
@@ -613,6 +613,10 @@ namespace DirectX
     constexpr unsigned long TEX_FILTER_MODE_MASK = 0xF00000;
     constexpr unsigned long TEX_FILTER_SRGB_MASK = 0xF000000;
 
+    // For reporting the image compression/conversion progress
```
walbourn, line 616: Callbacks should use ``std::function`` instead of legacy C style callbacks.

```
 _In_opt_ std::function<bool __cdecl(size_t, size_t)> statusCallBack = nullptr);
```

See the WIC functions for examples as well as [UVAtlas](https://github.com/microsoft/UVAtlas/blob/main/UVAtlas/inc/UVAtlas.h).

DirectXTex/DirectXTexP.h
```diff
@@ -199,6 +199,9 @@ using WICPixelFormatGUID = GUID;
 // HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)
 #define HRESULT_E_NOT_SUPPORTED static_cast<HRESULT>(0x80070032L)
 
+// HRESULT_FROM_WIN32(ERROR_CANCELLED)
```
walbourn, line 202: Instead of using ``HRESULT_FROM_WIN32(ERROR_CANCELLED)``, just use ``E_ABORT``.

DirectXTex/DirectXTex.h
walbourn: A subtle but important detail: The public interface passes the ``std::function<>`` by value. All the internal functions take it by ``const std::function<>&``.

walbourn (changes requested):

DirectXTex/BCDirectCompute.cpp
0xC0000054: No, I have not. IIRC, I added that callback for parity with the CPU-based compressor code. 
The callback is only used by `HRESULT DirectX::CompressEx(ID3D11Device* pDevice,
    const Image& srcImage,
    DXGI_FORMAT format,
    TEX_COMPRESS_FLAGS compress,
    float alphaWeight,
    ScratchImage& image,
    ProgressProc progressProc) noexcept`, I could remove it and change that method to report 0% and 100%.

0xC0000054: I removed the progress reporting code from the DirectComput inner loop.

DirectXTex/DirectXTex.h
0xC0000054: Done.

DirectXTex/DirectXTexP.h
0xC0000054: Done.

DirectXTex/DirectXTex.h
```diff
@@ -632,6 +632,14 @@ namespace DirectX
     HRESULT __cdecl Convert(
         _In_reads_(nimages) const Image* srcImages, _In_ size_t nimages, _In_ const TexMetadata& metadata,
         _In_ DXGI_FORMAT format, _In_ TEX_FILTER_FLAGS filter, _In_ float threshold, _Out_ ScratchImage& result) noexcept;
+
+    HRESULT __cdecl ConvertEx(
+        _In_ const Image& srcImage, _In_ DXGI_FORMAT format, _In_ TEX_FILTER_FLAGS filter, _In_ float threshold,
+        _Out_ ScratchImage& image, _In_opt_ std::function<bool __cdecl(size_t, size_t)> statusCallBack = nullptr) noexcept;
```
walbourn, line 638: For correctness, we probably need to remove ``noexcept`` from the specification for the Ex versions since we can't make any guarantees about the callback.

walbourn:
Since we are adding a new signature here, how about we make the `CompressEx` versions take a struct instead of as individual things like threshold values?

```
struct CompressOptions
{
    TEX_COMPRESS_FLAGS flags;
    float threshold;
    float alphaWeight;
};

HRESULT __cdecl CompressEx(
        _In_ const Image& srcImage, _In_ DXGI_FORMAT format, _In_ const CompressOptions& options,
        _Out_ ScratchImage& cImage, _In_opt_ std::function<bool __cdecl(size_t, size_t)> statusCallBack = nullptr) noexcept;
```

This would let me add new options to the Ex versions without breaking the ABI.

Just a thought.

DirectXTex/DirectXTex.h
0xC0000054: Done.

0xC0000054:
Should the struct have a version number/size field for compatibility between different versions of DirectXTex?
Or is this not a concern due to most callers using static linking.

walbourn:
> Should the struct have a version number/size field for compatibility between different versions of DirectXTex? Or is this not a concern due to most callers using static linking.

No need as you note.

DirectXTex/DirectXTex.h
```diff
@@ -632,6 +636,13 @@ namespace DirectX
     HRESULT __cdecl Convert(
         _In_reads_(nimages) const Image* srcImages, _In_ size_t nimages, _In_ const TexMetadata& metadata,
         _In_ DXGI_FORMAT format, _In_ TEX_FILTER_FLAGS filter, _In_ float threshold, _Out_ ScratchImage& result) noexcept;
+
+    HRESULT __cdecl ConvertEx(
+        _In_ const Image& srcImage, _In_ DXGI_FORMAT format, _In_ TEX_FILTER_FLAGS filter, _In_ float threshold,
+        _Out_ ScratchImage& image, _In_opt_ ProgressProc progressProc) noexcept;
```
walbourn: Same thing with ConvertEx. Remove 'noexcept'

walbourn:
Looks good.

**One final thought:** Should we also make the Ex convert version future-expandable in the same way?

```
struct ConvertOptions
{
    TEX_FILTER_FLAGS filter;
    float            threshold;
};

HRESULT __cdecl ConvertEx(
    _In_ const Image& srcImage, _In_ DXGI_FORMAT format, _In_ const ConvertOptions& options,
    _Out_ ScratchImage& image, _In_opt_ std::function<bool __cdecl(size_t, size_t)> statusCallBack = nullptr);
```

> It's always so hard to go back to update the ABI in future without breaking existing code.

0xC0000054:
> One final thought: Should we also make the Ex convert version future-expandable in the same way?

Done.

DirectXTex/DirectXTex.h
```diff
@@ -613,6 +613,12 @@ namespace DirectX
     constexpr unsigned long TEX_FILTER_MODE_MASK = 0xF00000;
     constexpr unsigned long TEX_FILTER_SRGB_MASK = 0xF000000;
 
+    struct ConvertOptions
```
walbourn, line 616: Please line up the fields in this struct like you did for ``CompressOptions``.

0xC0000054: Done.

walbourn (approved):

walbourn:
Thanks again for the work. I'm validating it with the test suite now and then I'll merge it.

walbourn:
Wiki page updates:

https://github.com/microsoft/DirectXTex/wiki/Convert

0xC0000054:
I think that the wiki should mention that the callback takes the form of `bool Progress(size_t done, size_t total)`.
It may not be not immediately obvious from the `std::function<bool (size_t, size_t)>` signature what the two `size_t` parameters represent.

walbourn:
@0xC0000054 Thanks again for your contribution. This work shipped in the [September 2023 release](https://github.com/microsoft/DirectXTex/releases/tag/sept2023).

PR #384, opened 2023-08-18
Add NO_WCHAR_T build option to CMake

walbourn:
Some usage scenarios build the library with ``/Zc:wchar_t-`` but mix both ``unsigned short*`` and ``wchar_t*`` client usage. For this *opt-in* build scenario, the library will link with both the non-native and native ``wchar_t``.

> Also fixes some Code Analysis warnings as you can't use ``_In_opt_`` with ``std::function<>``.


scm-xbox (approved):

walbourn:
Note that after completing this one, I'll need to make sure the same adapter is in place for the Xbox extensions.

PR #393, opened 2023-09-08
Support reading files with a DDS_PIXELFORMAT size of 24

### LINKED ISSUE #392: DDS files with a DDS_PIXELFORMAT size of 24
I encountered these files as part of a [bug report](https://forums.getpaint.net/topic/122998-cant-open-certain-dds-files-in-newer-version-but-can-in-3511/) on my Paint.NET DDS plugin.
The files are associated with mods for the Steam game Flat Out 2, all files use the legacy DX 9 header with DXT1-DXT5.

Whatever tool produced these files correctly sets the DDS header size, but for some reason it sets the `DDS_PIXELFORMAT` size to 24 instead of 32.

[ddspf_24.zip](https://github.com/microsoft/DirectXTex/files/12551254/ddspf_24.zip)

0xC0000054:
This is a common issue with the legacy DX 9 textures used by the game FlatOut 2.
The other DDS header fields are correct, so it is probably a bug in the program that created those files.

Fixes #392

### DIFF
--- a/DirectXTex/DirectXTexDDS.cpp
+++ b/DirectXTex/DirectXTexDDS.cpp
@@ -363,6 +363,7 @@ namespace
         {
             if (pHeader->ddspf.size != sizeof(DDS_PIXELFORMAT))
             {
+                // We do not accept legacy DX9 'known variants' for modern "DX10" extension header files.
                 return E_FAIL;
             }
 
--- a/DirectXTex/DirectXTexDDS.cpp
+++ b/DirectXTex/DirectXTexDDS.cpp
@@ -346,7 +346,9 @@ namespace
             return E_FAIL;
         }
 
-        if (pHeader->ddspf.size != 0 && pHeader->ddspf.size != sizeof(DDS_PIXELFORMAT))
+        if (pHeader->ddspf.size != 0 /* Known variant */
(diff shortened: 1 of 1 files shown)


DirectXTex/DirectXTexDDS.cpp
```diff
@@ -359,6 +361,11 @@ namespace
         if ((pHeader->ddspf.flags & DDS_FOURCC)
             && (MAKEFOURCC('D', 'X', '1', '0') == pHeader->ddspf.fourCC))
         {
+            if (pHeader->ddspf.size != sizeof(DDS_PIXELFORMAT))
+            {
+                return E_FAIL;
```
walbourn, line 366: Maybe add comment here:

```
if (pHeader->ddspf.size != sizeof(DDS_PIXELFORMAT))
{
    // We do not accept legacy DX9 'known variants' for modern "DX10" extension header files.
    return E_FAIL;
}
```

0xC0000054: Done.

walbourn:
I'm going to accept this, but will modify it to only accept the DDPF.dwSize values of '0' or '24' if a new ``DDS_FLAGS_PERMISSIVE`` is provided to the reader.

walbourn (approved):

PR #394, opened 2023-09-11
DDS_FLAGS_PERMISSIVE flag to allow some file variants with header bugs

walbourn:
In #371 the DDS reader was updated to allow the header to contain a DDPIXELFORMAT.dwSize of 0 and a valid FourCC code.

In #393 the DDS reader was updated to allow the header to contain a DDPIXELFORMAT.dwSize of 24 as long as the DDS_HEADER.dwSize size was correct.

This moves both of those cases to require the ``DDS_FLAGS_PERMISSIVE`` flag or the reader will reject them as invalid DDS files which was the original behavior.

Furthermore, with the ``DDS_FLAGS_PERMISSIVE`` flag the reader will also accept DDS_HEADERs with mipCounts that are too large for the input size. These will be bounded to the *correct* value. The most common file case here is a mipcount that is one too large, but this handles any too large mipCount.

The **texconv** tool now accepts ``-permissive`` as a command-line parameter for these cases.

> All of these cases are only supported for legacy DX9 DDS Files. Modern "DX10" header extension files are required to write correct values in the header to be considered valid.

### DIFF
--- a/DirectXTex/DirectXTex.h
+++ b/DirectXTex/DirectXTex.h
@@ -226,6 +226,9 @@ namespace DirectX
         DDS_FLAGS_BAD_DXTN_TAILS = 0x40,
         // Some older DXTn DDS files incorrectly handle mipchain tails for blocks smaller than 4x4
 
+        DDS_FLAGS_PERMISSIVE = 0x80,
+        // Allow some file variants due to common bugs in the header written by various leagcy DDS writers
+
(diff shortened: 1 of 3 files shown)


walbourn:
@0xC0000054 Please take a look...

walbourn:
Test suite updates in [this PR](https://github.com/walbourn/directxtextest/pull/30).

scm-xbox (approved):

0xC0000054:
Looks good to me.

PR #395, opened 2023-09-14
added shared build option

bmarques1995:


walbourn:
The library does not support DLL import/export. There are no decorations in place to enable that, so just changing to a ``SHARED`` in the CMake is not sufficient.

walbourn:
Generally speaking, I do not recommend just making DirectXTex a DLL in any case. Write some kind of wrapper interface that uses a stable C or COM ABI, export the functionality from your DLL, and link it to the DirectXTex library statically.

PR #400, opened 2023-10-03
Read legacy files with a DDS_HEADER size of 24

### LINKED ISSUE #399: Legacy DDS files with a DDS_HEADER size of 24
Related to https://github.com/0xC0000054/pdn-ddsfiletype-plus/issues/16

The OP says the files are from S.T.A.L.K.E.R.: Shadow of Chernobyl, a DX 9-based game.
I have no idea why these files use 24 for the DDS_HEADER size, did they forget the 1 when typing 124.

A sample for your test suite: [dds_header size 24.zip](https://github.com/microsoft/DirectXTex/files/12797485/dds_header.size.24.zip)

I plan to submit a PR with the following fix. and an updated check in the DX 10 code path:

```cpp
if (flags & DDS_FLAGS_PERMISSIVE)
{
    if (pHeader->size != 24 /* Known variant */
        && pHeader->size != sizeof(DDS_HEADER))
    {
        return HRESULT_E_NOT_SUPPORTED;
    }
}
else if (pHeader->size != sizeof(DDS_HEADER))
{
    return HRESULT_E_NOT_SUPPORTED;
}
```

0xC0000054:
Some DDS writers incorrectly use the value of *24* instead of *124* for the size of the ``DDS_HEADER``. Otherwise the file is valid. This PR updates the DDS reader to accept these files if the ``DDS_FLAGS_PERMISSIVE`` flag is used.

For "DX10" extended header scenarios, the header size must always be **124**.

### DIFF
--- a/DirectXTex/DirectXTexDDS.cpp
+++ b/DirectXTex/DirectXTexDDS.cpp
@@ -341,9 +341,17 @@ namespace
         auto pHeader = reinterpret_cast<const DDS_HEADER*>(static_cast<const uint8_t*>(pSource) + sizeof(uint32_t));
 
         // Verify header to validate DDS file
-        if (pHeader->size != sizeof(DDS_HEADER))
+        if (flags & DDS_FLAGS_PERMISSIVE)
         {
-            return E_FAIL;
+            if (pHeader->size != 24 /* Known variant */
+                && pHeader->size != sizeof(DDS_HEADER))
+            {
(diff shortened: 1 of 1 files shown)


walbourn (approved):

walbourn:
Test suite updates in [this commit](https://github.com/walbourn/directxtextest/commit/6fb69f84c3fd81ac04d10a38581a68eb7133d807).

Issue #426, opened 2023-12-24
Does DirectXTex directly depend on ZLIB?


JPeterMugaas:
I ask because Auxiliary
/DirectXTexEXR.cpp line 25: states

`// Requires the OpenEXR library <http://www.openexr.com/> and ZLIB <http://www.zlib.net>`

but the only mention in CMakeLists.txt is line 352 that adds it to the pkg-config file.  I do not see anything such as:

`find_package(ZLIB)`

If DirectXTex directly requires ZLib, then that DirectXKub;s CNajeKusts should link the library with ZLIB  with something like `target_link_libraries(${PROJECT_NAME} ZLIB::ZLIB)` after finding the ZLIB package.  

If this library does not directly require ZLIB,, CMakeLists.txt should NOT add ZLIB as a dependency in the pkg-config file and the comment I mentioned needs to be fixed and ZLIB should NOT be listed as required in its pkg-config file in this library.

I will look at this further later today.

walbourn:
ZLIB is a dependency of OpenEXR.

JPeterMugaas:
We are talking about an indirect dependency, NOT a direct dependency.  It seems to me that zlib should be removed from the removed from the requires field in the pkg-config file for this package.  If OpenEXR depends upon ZLIB, then that should have been listed in its OpenEXR.pc file.

PR #427, opened 2023-12-25
Remove references to ZLIB and edit comment as appropriate.

JPeterMugaas:
Fixes https://github.com/microsoft/DirectXTex/issues/426 .

### DIFF
--- a/Auxiliary/DirectXTexEXR.cpp
+++ b/Auxiliary/DirectXTexEXR.cpp
@@ -22,7 +22,7 @@
 #include <tuple>
 
 //
-// Requires the OpenEXR library <http://www.openexr.com/> and ZLIB <http://www.zlib.net>
+// Requires the OpenEXR library <http://www.openexr.com/> and its dependencies.
 //
 
 #ifdef __clang__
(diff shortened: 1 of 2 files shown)


JPeterMugaas:
I have a few comments:

First, ArchLinux (https://archlinux.org/packages/extra/x86_64/libjpeg-turbo/) and the MSYS2 (https://www.msys2.org/) mingw-w64 distributions use libjpeg-turbo (https://libjpeg-turbo.virtualgl.org/) .  That does NOT have the `JCS_BG_RGB` constant so you need to ifdef it with something such as `#ifdef JCS_BG_RGB` in your case statement at line 74 in DirectXTexJPEG.cpp.

Second, I get the following warning with GCC 13.2.0:
```
C:/msys64/home/jpmugaas/directxtex/Auxiliary/DirectXTexPNG.cpp: In function 'void DirectX::OnPNGWarning(png_structp, png_const_charp)':
C:/msys64/home/jpmugaas/directxtex/Auxiliary/DirectXTexPNG.cpp:33:5: warning: 'noreturn' function does return
   33 |     }
      |     ^
```
You should remove. the "`[[noreturn]]`" from line 30 in Auxiliary/DirectXTexPNG.cpp to fix this.

and

Third, pkg-config support was recently added so you need to add the following lines in CMakeLists at around line 390:
```
if(JPEG_FOUND)
  list(APPEND DIRECTXTEX_DEP_L "libjpeg")
endif()
if(PNG_FOUND)
  list(APPEND DIRECTXTEX_DEP_L "libpng")
endif()
if(ZLIB_FOUND)
  list(APPEND DIRECTXTEX_DEP_L "zlib")
endif()
```
This will help some users handle the direct dependencies.

walbourn (approved):

PR #431, opened 2023-12-30
December 31, 2023

walbourn:
* TGA reader bug fix for offset boundary check
* PPM/PFM reader bug fixes for bounds checking
* Xbox-specific functionality added to Auxiliary folder
* CMake project updates including pkg-config file generation
* Code review


JPeterMugaas:
You need to also release an updated DirectXMath because the pkg-config file lists that library as a requirement.

walbourn:
> You need to also release an updated DirectXMath because the pkg-config file lists that library as a requirement.

See https://github.com/microsoft/DirectXMath/pull/175

scm-xbox (approved):

PR #438, opened 2024-01-10
CMake update to install EXEs

JPeterMugaas:


walbourn:
Why doesn't yhe existing 'cmake --install' not work here?

CMakeLists.txt
```diff
@@ -668,6 +668,17 @@ if(BUILD_TOOLS AND BUILD_DX11 AND WIN32)
   set_property(DIRECTORY PROPERTY VS_STARTUP_PROJECT texconv)
 endif()
 
+#--- install .exes
+
+if(MINGW AND BUILD_TOOLS)
+    foreach(t IN LISTS TOOL_EXES ITEMS ${PROJECT_NAME})
+      install(TARGETS ${t}
+         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
walbourn, line 676: For exe files, is archive or library a required element?

CMakeLists.txt
```diff
@@ -668,6 +668,17 @@ if(BUILD_TOOLS AND BUILD_DX11 AND WIN32)
   set_property(DIRECTORY PROPERTY VS_STARTUP_PROJECT texconv)
 endif()
 
+#--- install .exes
```
walbourn, line 671: Remove this comment plz

CMakeLists.txt
```diff
@@ -668,6 +668,17 @@ if(BUILD_TOOLS AND BUILD_DX11 AND WIN32)
...
+
+if(MINGW AND BUILD_TOOLS)
```
walbourn, line 673: This is not particularly specific to MinGW. That said, I have to handle the EXE publishing for VCPKG in a special way to use the code-signed versions when possible.

Please make this:

```
if(BUILD_TOOLS AND (NOT VCPKG_TOOLCHAIN))
```

CMakeLists.txt
```diff
@@ -668,6 +668,17 @@ if(BUILD_TOOLS AND BUILD_DX11 AND WIN32)
...
+    foreach(t IN LISTS TOOL_EXES ITEMS ${PROJECT_NAME})
```
walbourn, line 674: Remove the ``ITEMS ${PROJECT_NAME}`` since that install logic is already handled above.

CMakeLists.txt
walbourn: ``RUNTIME DESTINATION`` is all you need for these.

PR #447, opened 2024-01-30
Add ability to build with DX11 support including shaders in MINGW.

JPeterMugaas:
This uses a wine-compatible (fxc2) program available in MSYS2.  The program home page is: # https://github.com/mozilla/fxc2/ .


CMakeLists.txt
```diff
@@ -214,17 +239,33 @@ if(BUILD_DX11 AND WIN32)
...
-        add_custom_command(
-            OUTPUT "${COMPILED_SHADERS}/BC6HEncode_EncodeBlockCS.inc"
-            MAIN_DEPENDENCY "${PROJECT_SOURCE_DIR}/DirectXTex/Shaders/CompileShaders.cmd"
-            DEPENDS ${SHADER_SOURCES}
-            COMMENT "Generating HLSL shaders..."
-            COMMAND ${CMAKE_COMMAND} -E env CompileShadersOutput="${COMPILED_SHADERS}" CompileShaders.cmd > "${COMPILED_SHADERS}/compileshaders.log"
-            WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/DirectXTex/Shaders"
-            USES_TERMINAL)
+	    if(MINGW)
+		    message(STATUS "Generating HLSL shaders...")
```
walbourn, line 244: Why not just update the CompilerShaders.cmd script directly?

Where is your 'fxc' coming from in this case?

walbourn: Note the script should default to just 'fxc.exe' if it can't find the Windows SDK locations through environment variables already

JPeterMugaas: > Why not just update the CompilerShaders.cmd script directly?

I did install the latest Windows SDK but that does not have a command-line, and it can be difficult to access this from the MSYS2 environment particularly as the paths contain spaces.  I did base the CMake macros partly on the .cmd script although.  I also have to be honest in saying that I probably am more familiar with CMake then the .cmd command-lines shell.  Besides, I'm not sure I like having to invoke CMake and a cmd shell.  It's not as flexible as I required.

> Where is your 'fxc' coming from in this case?

The fxc program I obtained is downloadable from the MSYS2  ( https://www.msys2.org/ ) distribution in a pacman-package and the home page for the program is https://github.com/mozilla/fxc2 .

JPeterMugaas: Perhaps, it may help if I explain what I'm trying to do.  I'm writing a PKGBUILD script that builds DirectXTex into a redistributable package that I contribute to the MSYS2 ( http://www.msys2.org ) project.  I can't count on the computer building this redistributable package to have the Windows SDK installed along with the global path.  MSYS2 tries to be a self-contained Unix-like environment for building Windows software.  I was able to use the changes to CMakeLists.txt to build DirectXTex with Directx11 run-time support and even went so far as to build the ddsviewer.

walbourn: Are you saying the wine FXC is always present in MSYS2?

JPeterMugaas: There is a fxc.exe  but that is not always installed.  It can be installed simply by using "pacman" to install the appropriate fxc2 packages (this is strictly OPTIONAL).  What I will probably do is add "fxc2" as a buildtime dependency meaning it would be required to build DirectXTex package with DirectX 11 support.

CMakeLists.txt
```diff
@@ -203,6 +203,34 @@ if(ENABLE_LIBPNG_SUPPORT)
     list(APPEND LIBRARY_SOURCES Auxiliary/DirectXTexPNG.cpp)
 endif()
 
+macro(call_mingw_fxc work_dir input_filename output_filename entry_point var_name semantic _targetname)
```
walbourn, line 206: Why not put these macros into your .cmake?

CMakeLists.txt
```diff
@@ -214,17 +242,31 @@ if(BUILD_DX11 AND WIN32)
...
-        add_custom_command(
-            OUTPUT "${COMPILED_SHADERS}/BC6HEncode_EncodeBlockCS.inc"
-            MAIN_DEPENDENCY "${PROJECT_SOURCE_DIR}/DirectXTex/Shaders/CompileShaders.cmd"
-            DEPENDS ${SHADER_SOURCES}
-            COMMENT "Generating HLSL shaders..."
-            COMMAND ${CMAKE_COMMAND} -E env CompileShadersOutput="${COMPILED_SHADERS}" CompileShaders.cmd > "${COMPILED_SHADERS}/compileshaders.log"
-            WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/DirectXTex/Shaders"
-            USES_TERMINAL)
+	    if(MINGW)
+		    file(COPY ${PROJECT_SOURCE_DIR}/DirectXTex/Shaders/BC7Encode.hlsl DESTINATION ${COMPILED_SHADERS})
```
walbourn, line 247: Why copy the shader source files instead of using them in place?

CMakeLists.txt
```diff
@@ -214,17 +242,31 @@ if(BUILD_DX11 AND WIN32)
...
+			file(COPY ${PROJECT_SOURCE_DIR}/DirectXTex/Shaders/BC6HEncode.hlsl DESTINATION ${COMPILED_SHADERS})
+		
+		    call_mingw_fxc_dx11(${COMPILED_SHADERS} BC7Encode TryMode456CS)
```
walbourn, line 250: If you are going to implement the shader compilation inside the CMake, why make it MinGW only? Just use it always for WIN32 & BUILD_DX11 and rename the .cmake and macros to something more generic: ``hlsl_compile`` perhaps?

> This would mean both the CMake and the .cmd script would have to be kept in sync, but at least for this library there are not that many shaders involved.

CMakeLists.txt
```diff
@@ -525,38 +567,58 @@ if(BUILD_TOOLS AND WIN32)
...
+	  "${var_name}" "${semantic}" "${_fname}")
+  list(APPEND DDS_SRC_L "${_fname}")
+endmacro()
+  
   if(NOT COMPILED_DDSVIEW_SHADERS)
       if(USE_PREBUILT_SHADERS)
           message(FATAL_ERROR "ERROR: Using prebuilt shaders requires the COMPILED_DDSVIEW_SHADERS variable is set")
       endif()
-      set(COMPILED_DDSVIEW_SHADERS ${CMAKE_CURRENT_BINARY_DIR}/Shaders/Compiled)
+      set(COMPILED_DDSVIEW_SHADERS ${CMAKE_CURRENT_BINARY_DIR}/Shadersd/Compiled)
```
walbourn, line 589: There's no file name collisions here, so I don't think you need a distinct folder name here.

CMakeLists.txt
```diff
@@ -203,6 +203,34 @@ if(ENABLE_LIBPNG_SUPPORT)
     list(APPEND LIBRARY_SOURCES Auxiliary/DirectXTexPNG.cpp)
 endif()
 
+macro(call_mingw_fxc work_dir input_filename output_filename entry_point var_name semantic _targetname)
+    add_custom_command(
+        COMMAND ${CMAKE_COMMAND} -DWORK_DIR=${work_dir}
```
walbourn, line 208: General note: Please make sure you convert all TABs to SPACES and TRIM WHITESPACE.

DDSView/ddsview.cpp
```diff
@@ -519,12 +523,12 @@ HRESULT InitDevice(const TexMetadata& mdata)
     case TEX_DIMENSION_TEXTURE1D:
         if (mdata.arraySize > 1)
         {
-            pshader = g_PS_1DArray;
+            pshader = (BYTE*) g_PS_1DArray;
```
walbourn, line 526: Prefer to use C++ casting either ``static_cast`` or ``reinterpret_cast``.

walbourn (changes requested):
I'm still debating taking this, but I can see value in having the CMake not rely on the DOS script files...

> I'd be less inclined to take this kind of change for DirectXTK/DirectXTK12 because of the myriad shader combinations and Xbox-specific codepaths, but for DirectXTex there isn't nearly as much to keep in sync between CMake and .cmd.

CMakeLists.txt
JPeterMugaas: Because we have to track what files are built so we can setup the appropriate dependencies.   The macros just save a little typing.

CMakeLists.txt
JPeterMugaas: fxc gave a sigmentat fault when I tried using them in place.

CMakeLists.txt
JPeterMugaas: It's mingw only because I don't have Visual Studio and I would not the where-with-all to use it in a timely manner even if I got a copy today.

CMakeLists.txt
JPeterMugaas: typo.

CMakeLists.txt
JPeterMugaas: I have updated the script to support Microsoft's SDK fxc.exe.  I was able to finally test it.  The recent change also calls the CMake script, hlsl_compile.cmake even in non-MINGW environments.  You better test MSVC to be sure that I didn't miss anything.

DDSView/ddsview.cpp
JPeterMugaas: I did that typecase because I was getting this error message:

```
:/msys64/home/jpmugaas/directxtex-PKGBUILD/src/directxtex-dec2023/DDSView/ddsview.cpp:522:23: error: invalid conversion from 'const signed char*' to 'const BYTE*' {aka 'const unsigned char*'} [-fpermissive]
  522 |             pshader = g_PS_1DArray;
      |                       ^~~~~~~~~~~~
      |                       |
      |                       const signed char*

```

walbourn:
```
pshader = reinterpret_cast<const BYTE*>(g_PS_1DArray);
```

CMakeLists.txt
JPeterMugaas: I also think the .cmd script was depending upon being in the same dir as the source directory..  That is not the case with the CMakeLists.txt file so we needed to copy the source-file to get things to work properly since the source .hlsl is a dependency for the target compiled .inc file.  I hope I'm making sense.  In any case, I see this as a minor point  that may not be working "fixing".

CMakeLists.txt
JPeterMugaas: I have gone ahead and modified the hlsl_compile so that it will try to use Microsoft's fxc syntax and if that fails, then try the subset for fxc2.

walbourn:
Please rebase (not merge) against main.

JPeterMugaas:
I've reevaluated the entire situation and I have another solution to my issue.

PR #471, opened 2024-05-01
Fix for HDR codec to avoid buffer overread for some malformed files

walbourn:
Fixed an issue when decoding the HDR text header if there is no ending `\n` present.


scm-xbox (approved):

walbourn:
Test suite updates here: https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/10687605

PR #483, opened 2024-06-23
Add cubemap support for OpenEXR

vkaytsanov:
OpenEXR format can includes metadata indicating the image is cubemap which is six images in a single file.

### DIFF
--- a/Auxiliary/DirectXTexEXR.cpp
+++ b/Auxiliary/DirectXTexEXR.cpp
@@ -373,7 +373,7 @@ HRESULT DirectX::LoadFromEXRFile(const wchar_t* szFile, TexMetadata* metadata, S
 
         int width = dw.max.x - dw.min.x + 1;
         int height = dw.max.y - dw.min.y + 1;
-        bool isEnvmap = false;
+        size_t arraySize = 1;
 
         if (width < 1 || height < 1)
             return E_FAIL;
@@ -384,28 +384,21 @@ HRESULT DirectX::LoadFromEXRFile(const wchar_t* szFile, TexMetadata* metadata, S
             {
                 height = width;
             }
-            isEnvmap = true;
+            arraySize = 6;
         }
 
         if (metadata)
         {
             metadata->width = static_cast<size_t>(width);
             metadata->height = static_cast<size_t>(height);
-            metadata->depth = metadata->arraySize = metadata->mipLevels = 1;
+            metadata->depth = metadata->mipLevels = 1;
+            metadata->arraySize = arraySize;
             metadata->format = DXGI_FORMAT_R16G16B16A16_FLOAT;
             metadata->dimension = TEX_DIMENSION_TEXTURE2D;
         }
 
-        if (!isEnvmap)
-        {
-            hr = image.Initialize2D(DXGI_FORMAT_R16G16B16A16_FLOAT,
-                static_cast<size_t>(width), static_cast<size_t>(height), 1u, 1u);
-        }
-        else
-        {
-            hr = image.InitializeCube(DXGI_FORMAT_R16G16B16A16_FLOAT,
-                static_cast<size_t>(width), static_cast<size_t>(height), 1u, 1u);
-        }
+        hr = image.Initialize2D(DXGI_FORMAT_R16G16B16A16_FLOAT,
+                static_cast<size_t>(width), static_cast<size_t>(height), arraySize, 1u);
         
         if (FAILED(hr))
             return hr;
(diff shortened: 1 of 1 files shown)


vkaytsanov:

walbourn:
Thanks. I didn't realize OpenEXR stored envmaps directly. Can you point me to an example data file?

vkaytsanov:
Hello, the test I did was using their tool that ships with vcpkg, it is called `exrenvmap.exe`. Official documentation: https://openexr.com/en/latest/bin/exrenvmap.html

These are the test files that the tool used and produced `skybox.exr` in the archive. Note: I could only open these files with Photoshop to view them:
[forest.zip](https://github.com/user-attachments/files/16042112/forest.zip)

walbourn:
For ``texassemble`` it makes sense to load an `.EXR` with "envmap" as an array of six images and write it as a cube-map or array texture. For ``texconv`` it probably should be treated as a texture array for consistent behavior within the tool.

Auxiliary/DirectXTexEXR.cpp
```diff
@@ -386,8 +396,17 @@ HRESULT DirectX::LoadFromEXRFile(const wchar_t* szFile, TexMetadata* metadata, S
...
-        hr = image.Initialize2D(DXGI_FORMAT_R16G16B16A16_FLOAT,
-            static_cast<size_t>(width), static_cast<size_t>(height), 1u, 1u);
+        if (!isEnvmap)
+        {
+            hr = image.Initialize2D(DXGI_FORMAT_R16G16B16A16_FLOAT,
+                static_cast<size_t>(width), static_cast<size_t>(height), 1u, 1u);
+        }
+        else
+        {
+            hr = image.InitializeCube(DXGI_FORMAT_R16G16B16A16_FLOAT,
```
walbourn, line 406: Within the context of the command-line tools, it should probably be an array and not a cube. I can add behavior flags to allow it to do either, or I can just make it always load as a texture array. I can then add EXR support to ``texassemble`` to support EXR to cubemap DDS conversions.

walbourn: IOW: I suggest this code be:

```
hr = image.Initialize2D(DXGI_FORMAT_R16G16B16A16_FLOAT,
            static_cast<size_t>(width), static_cast<size_t>(height), (isEnvmap) ? 6u : 1u, 1u);
```

vkaytsanov: Fair enough, updated. Keep in mind that the code was tested before the final changes as loading and initializing cubemap texture for a DXR Miss Shader

Auxiliary/DirectXTexEXR.cpp
```diff
@@ -371,23 +371,35 @@ HRESULT DirectX::LoadFromEXRFile(const wchar_t* szFile, TexMetadata* metadata, S
...
+        int width = dw.max.x - dw.min.x + 1;
+        int height = dw.max.y - dw.min.y + 1;
+        size_t arraySize = 1;
 
         if (width < 1 || height < 1)
             return E_FAIL;
 
+        if (file.header().find("envmap") != file.header().end())
+        {
+            if (width == height / 6)
```
walbourn, line 383: This is a good fix, but I wonder if the `arraySize = 6;` should be in the `(width == height / 6)` case to just ignore the env map thing if the size isn't right.

walbourn: I.e:

```
if (width == height / 6)
{
    height = width;
    arraySize = 6;
}
```

Auxiliary/DirectXTexEXR.cpp
```diff
@@ -371,23 +371,35 @@ HRESULT DirectX::LoadFromEXRFile(const wchar_t* szFile, TexMetadata* metadata, S
...
+            arraySize = 6;
+        }
+
         if (metadata)
         {
             metadata->width = static_cast<size_t>(width);
             metadata->height = static_cast<size_t>(height);
-            metadata->depth = metadata->arraySize = metadata->mipLevels = 1;
+            metadata->depth = metadata->mipLevels = 1;
+            metadata->arraySize = arraySize;
```
walbourn, line 395: `GetMetadataFromEXRFile` needs to match this behavior as well.

walbourn:
I'll take this PR and do the code review in a new PR.

Thanks!

walbourn (approved):

Auxiliary/DirectXTexEXR.cpp
vkaytsanov: Maybe it will be better if the size isn't right to return `E_FAIL`. Per their documentation it says it should contain the flag `envmap`

PR #484, opened 2024-06-27
texconv: -ignoremips options for reading some invalid DDS files

walbourn:
Some DDS files have invalid mipcount values for the amount of data provided. While ``-permissive`` already handles cases where the mipcount value in the header is "too big" for the size of the texture, this new ``-ignoremips`` option allows files that are simply missing data by only reading the top-level mip.

Also added support to **texdiag** for ``-permissive`` and ``-ignoremips`` (it already supported `-badtails`).


scm-xbox (approved):

walbourn:
Test coverage in [this PR](https://github.com/walbourn/directxtextest/pull/41).

PR #490, opened 2024-07-19
Some DDS DX10 writers set arraySize to 0 instead of 1

walbourn:
The vast majority of "DX10" extended header DDS files are written by D3DX10/D3DX11 or DirectXTex, and properly set the ``DDS_HEADER_DXT10.arraySize`` field to 1 or greater. Some 3rd party writer is setting it to '0' instead for '1' (thinking since it wasn't an array, it doesn't need a value here), which was a common pattern for mipLevels set to 0 for legacy DX9 format DDS files as well.


walbourn:
Test coverage https://github.com/walbourn/directxtextest/pull/42

Andrew-Farrier (approved):

PR #501, opened 2024-08-29
Support loading legacy DDS files using mixed channel formats

walbourn:
The legacy DDS file formats ``D3DFMT_A2W10V10U10``, ``D3DFMT_L6V5U5``, and ``D3DFMT_X8L8V8U8`` were originally not implemented because no DXGI format has a mix of signed/unsigned channels the way these legacy Direct3D 9 formats were made (the L and A channels are unsigned, the UVW channels are signed). They are also tied to legacy "BumpMap" technology, so generally weren't not considered that relevant to modern Direct3D.

There is a customer request that the library support loading these formats. For this PR, the unsigned values are remapped to the unorm range through biasing the value (i.e. effectively the same as x2 bias, but not applied to the L or A channels).


walbourn:
Test suite updates in [this PR](https://github.com/walbourn/directxtextest/pull/45)

PR #503, opened 2024-09-02
Fix exporting R8 files as PNG using libpng

brunoanc:
Fixes an incorrect stride value for R8 images being exported to PNG on the experimental libpng SaveToPNGFile implementation.

Also updated to make it clear interlaced images are not supported.

### DIFF
--- a/Auxiliary/DirectXTexPNG.cpp
+++ b/Auxiliary/DirectXTexPNG.cpp
@@ -131,7 +131,7 @@ namespace
             if (interlacing != PNG_INTERLACE_NONE)
             {
                 throw std::invalid_argument{ "interlacing not supported" };
-            }           
+            }
             // color handling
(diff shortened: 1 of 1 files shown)


walbourn:
@luncliff Do you agree with this fix?

walbourn:
I add a formal test for this code, and verified this is the right fix.

luncliff (approved):
I mostly worked with reading/writing colored images.
Nice fix!

walbourn:
The CLA was triggered due to my edits.

@microsoft-github-policy-service agree company="Microsoft"

walbourn:
Updating test suite to cover this code explicitly https://github.com/walbourn/directxtextest/pull/46

walbourn (approved):
Thanks for the fix. Added a few more after testing.

PR #522, opened 2024-10-03
Add GitHub Action for BVTs with subset of tests

walbourn:
This builds the test suite such that it only needs a few test data files which are included directly in the test repo. This avoids LFS bandwidth traffic for the full set of test data and provides an automated BVT.

This PR makes sure ``assert`` macros in the library built for fuzz-testing remain in place even in release builds.

> Includes ADO YAML code review to clean up use of PATs, and to convert all use of vcpkg in the pipelines to 'manifest mode'.

> Also sets the dependabot to not spam quite so often for GHA.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,4 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: daily
+      interval: weekly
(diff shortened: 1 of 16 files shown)


walbourn:
https://github.com/walbourn/directxtextest/pull/48

scm-xbox (approved):

PR #525, opened 2024-10-06
Update DirectXTexXboxDDS.cpp

karurung:
Using const uint8_t* instead of const void* enhances type safety and improves code security.


walbourn:
All the "frommemory" functions take ``const void*`` pointers for the data blob. Perhaps you should update all the instances in `DirectXTex.h` as well as change ``Blob``.

walbourn:
There is also a non-Xbox version of ``EncodeDDSHeader``

karurung:
Ok, I'll try it

PR #526, opened 2024-10-06
Update DirectXTexXbox.h

karurung:
'void*' to 'uint8_t*'

### DIFF
--- a/Auxiliary/DirectXTexXbox.h
+++ b/Auxiliary/DirectXTexXbox.h
@@ -95,13 +95,13 @@ namespace Xbox
     // Image I/O
 
     HRESULT __cdecl GetMetadataFromDDSMemory(
-        _In_reads_bytes_(size) const void* pSource, _In_ size_t size,
+        _In_reads_bytes_(size) const uint8_t* pSource, _In_ size_t size,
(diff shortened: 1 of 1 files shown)


walbourn:
Having this as distinct PRs is a little problematic because they won't build due to the mismatch of the h and cpp. Can you combine the header and .cpp files? If not, I can approve them in bulk after local testing.

walbourn:
Merging these changes into https://github.com/microsoft/DirectXTex/pull/538

Auxiliary/DirectXTexXbox.h
```diff
@@ -163,16 +163,16 @@ namespace Xbox
 
     HRESULT __cdecl CreateTexture(
         _In_ ID3D12Device* d3dDevice,
-        _In_ const XboxImage& xbox, _Outptr_opt_ ID3D12Resource** ppResource, _Outptr_ void** grfxMemory);
+        _In_ const XboxImage& xbox, _Outptr_opt_ ID3D12Resource** ppResource, _Outptr_ const uint8_t** grfxMemory);
```
walbourn, line 166: Note ``void*`` is actually the correct annotation for graphics memory. This is untyped memory in the video device range so could be really anything and maybe not byte accessible.

PR #532, opened 2024-10-12
Fix DX12 Capture transition state handling for MSAA

walbourn:
Also updates ScreenGrab and DDSTextureLoader for latest changes.


walbourn:
Verified fixed with

https://github.com/walbourn/directxtextest/pull/53

scm-xbox (approved):

PR #538, opened 2024-10-24
Use fixed width integers for underlying enum types

walbourn:
Originally I used `DWORD` for flags. When I removed Windows-specific types, I changed it to `unsigned long` to keep the signature the same. The ask is to use `uint32_t` instead which is more consistent on Windows vs. Linux.

Also removed 'extra-semi' for macro.

Changed use of `void*` to `uint8_t*` for "FromMemory" operations which is the recommendation for C++ byte-addressable buffers. This matches how DDSTextureLoader/WICTextureLoader already do it.

Added a `GetConstBufferPointer` method to `Blob` utility class for const-correct access.


walbourn:
https://github.com/walbourn/directxtextest/pull/54

scm-xbox (approved):

PR #550, opened 2024-11-07
texconv: Portable Half Map uses .phm extension

walbourn:
I added support for the 'float16' variant of PFM some time back. ImageMagick uses a `.phm` extension for these files instead of `.pfm`.

> My PFM codec only supports reading "PHM" files, and not writing them so there's no need to add another 'CODEC'


walbourn:
https://github.com/microsoft/DirectXTex/commit/5999f34c1181cc8f69b2da796c638c2c7be72db3

scm-xbox (approved):

PR #555, opened 2024-11-12
texconv: add --ignore-srgb, --wic-uncompressed and revisit --wic-lossless

walbourn:
* Adds ``--ignore-srgb`` switch which passes `TGA_FLAGS_IGNORE_SRGB` or `WIC_FLAGS_IGNORE_SRGB` to the loader as appropriate.

* Adds ``--wic-uncompressed`` switch which applies to TIFF and HEIF.

* `--wic-lossless` should not be applied to JPEG as it only has a quality setting. This switch now only applies JPEG XR.

* Add `--wic-quality` handling for HEIF codec output.

* Remove ``heic`` as an option for output as the extension can read both, but only supports writing ``hief``.

### DIFF
--- a/Texconv/texconv.cpp
+++ b/Texconv/texconv.cpp
@@ -117,6 +117,7 @@ namespace
         OPT_TGAZEROALPHA,
         OPT_WIC_LOSSLESS,
         OPT_WIC_MULTIFRAME,
+        OPT_WIC_UNCOMPRESSED,
         OPT_NOLOGO,
         OPT_TIMING,
         OPT_SEPALPHA,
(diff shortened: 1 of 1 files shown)


scm-xbox (approved):

walbourn:
In my initial implementation of these switches, I was basing it on the 'general principle' that WIC has ImageQuality is between 0 and 1, and 1 means "lossless if you support it".

In practice over time, the tool only sets these on specific codecs so this has resulted in consumer confusion. In addition, I've conflated "no compression" with "lossless" in the case of TIFF.

This PR tries to straighten all that out.

walbourn:
https://github.com/walbourn/directxtextest/commit/765f2d3d40935eeff50f9b93e231b92889d0cb2d

PR #564, opened 2024-12-06
Build for a Windows 8.1 minimum

walbourn:
Windows 7 was 0x0601. Windows 8 was 0x0602. Both are well past their life-cycle. While Windows 8.1 is also out-of-support, it was only retired as of last year so I'm keeping it for now.

CMake presets like ``x64-Debug-Win7`` are now ``x64-Debug-Win8``.

Always uses CreateFile2 and WIC2


walbourn:
https://github.com/walbourn/directxtextest/pull/57

scm-xbox (approved):

Issue #572, opened 2025-01-29
Add support for build DirectXTex toolkit as shared libs (.dll)


bmarques1995:
I believe that will be useful this project add support for shared libs

bmarques1995:
Later I will study and think how to do this.

PR #573, opened 2025-01-29
Add option to build DXTex as shared lib to cmake

bmarques1995:
This pull request solves the #572, only for CMake, and doesn't modify the behavior of the previous build, tested, compiles for all utilities (ddsview, texconv, texassemble and texdiag)


walbourn:
I've always had these libraries focused on static C++ lib scenarios and not built as DLLs since there's no consistent ABI design (COM or C flat).

What's your scenario that a DLL is required?

Also, you could move all the declspec annotations to the class level instead of the individual methods for the non-free functions which would at least minimize the code-lines changed.

bmarques1995:
> I've always had these libraries focused on static C++ lib scenarios and not built as DLLs since there's no consistent ABI design (COM or C flat).
> 
> What's your scenario that a DLL is required?
> 
> Also, you could move all the declspec annotations to the class level instead of the individual methods for the non-free functions which would at least minimize the code-lines changed.

Hi Walbourn, my point to let the user decide if it want to use static or shared libs, taking my example, I'm developing a game engine, and, often i request new features and updates of packages, and since the DirectXTex is only static, at every change I need to redistribute the entire binary.

I believe u have seen that the default behavior was mantained (static linkage). Also, this change can turn the lib closer to KTX (OpenGL and Vulkan texture handler).

About your request to reduce the number of declspec, this is fine for me.

CMakeLists.txt
```diff
@@ -220,7 +222,12 @@ if(BUILD_DX11 AND WIN32 AND (NOT (XBOX_CONSOLE_TARGET STREQUAL "durango")))
     endif()
 endif()
 
-add_library(${PROJECT_NAME} STATIC ${LIBRARY_SOURCES} ${LIBRARY_HEADERS})
+add_library(${PROJECT_NAME} ${LIBRARY_SOURCES} ${LIBRARY_HEADERS})
```
walbourn, line 225: Doesn't this need to be ``STATIC`` for a C++ library and ``SHARED`` for a DLL?

walbourn: Oh, I see. This is a thing CMake deals with per https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS

DirectXTex/DirectXTex.h
```diff
@@ -49,30 +49,37 @@ struct IWICMetadataQueryReader;
 
 #define DIRECTX_TEX_VERSION 206
 
+#ifdef DXTEX_EXPORT
```
walbourn, line 52: For consistency with the other macro and to avoid any collision, perhaps you should use a longer name.

Might be nice to add ``__cdecl`` here as well to remove it from the rest of the header?

```
#ifdef DIRECTX_TEX_EXPORT
#define DIRECTX_TEX_API __declspec(dllexport) __cdecl
#elif defined(DIRECTX_TEX_IMPORT)
#define DIRECTX_TEX_API __declspec(dllimport) __cdecl
#else
#define DIRECTX_TEX_API __cdecl
#endif
```

bmarques1995: ok, this is fine for me

walbourn: Nevermind about the ``__cdecl`` thing. The placement isn't right.  Otherwise, we'd have to do something like ``DIRECTX_TEX_API(type) __declspec(dllimport) type __cdecl`` which is even uglier.

bmarques1995: > ok, this is fine for me

I tested the `__declspec` with the `__cdecl`, and if is the order, for example: `__declspec(dllexport) __cdecl constexpr bool  IsValid(_In_ DXGI_FORMAT fmt) noexcept;` is not allowed.

The compiler only accepts: `__declspec <return type> __cdecl <Body>()`

walbourn:
Required some test updates to get this to build.

https://github.com/walbourn/directxtextest/pull/58

bmarques1995:
> Required some test updates to get this to build.
> 
> [walbourn/directxtextest#58](https://github.com/walbourn/directxtextest/pull/58)

Walbourn, about the failed tests, do u have any idea about what is causing the `run-vcpkg` error? I'm failing in reproduce the error

bmarques1995:
> > Required some test updates to get this to build.
> > [walbourn/directxtextest#58](https://github.com/walbourn/directxtextest/pull/58)
> 
> Walbourn, about the failed tests, do u have any idea about what is causing the `run-vcpkg` error? I'm failing in reproduce the error

Forget this, Walbourn, I've fixed the workflows, now they are handling the latest vcpkg version

walbourn:
> About the previous reply, I saw that the modified workflows were not getting the correct vcpkg commit hash, so I added the step to get the hash and saved in the env, to later get it in `${{ env.VCPKG_COMMIT_ID}}`

I had the VCPKG-based workflows using a commitId value from the 'repository variables' and set them to the 'formal releases' rather than using 'latest HEAD', but workflows run by non-owners of the repo are restricted as secrets so that's why they were failing.

I wonder if I can set it up to use the var if accessible and fallback to "HEAD" otherwise.

walbourn:
I'm working on some additional commits here to add another build pipeline to validate this model and add `VERSIONINFO`... I should have it ready this weekend and we can wrap this PR up.

Thanks!

bmarques1995:
> > About the previous reply, I saw that the modified workflows were not getting the correct vcpkg commit hash, so I added the step to get the hash and saved in the env, to later get it in `${{ env.VCPKG_COMMIT_ID}}`
> 
> I had the VCPKG-based workflows using a commitId value from the 'repository variables' and set them to the 'formal releases' rather than using 'latest HEAD', but workflows run by non-owners of the repo are restricted as secrets so that's why they were failing.
> 
> I wonder if I can set it up to use the var if accessible and fallback to "HEAD" otherwise.

So, in this case, I only 

> I'm working on some additional commits here to add another build pipeline to validate this model and add `VERSIONINFO`... I should have it ready this weekend and we can wrap this PR up.
> 
> Thanks!

You're welcome.

.github/workflows/bvt.yml
```diff
@@ -81,11 +81,20 @@ jobs:
             echo "::error Unknown architecture/build-type triplet mapping"
         }
 
+    - name: Get latest vcpkg commit hash
```
walbourn, line 84: How about this version?

```
    - name: Get latest vcpkg commit hash
      shell: pwsh
      run: |
        if ($Env:vcpkgRelease) {
            echo "Using vcpkg commit from repo variable..."
            $VCPKG_COMMIT_ID = $Env:vcpkgRelease
        }
        else {
            echo "Fetching latest vcpkg commit hash..."
            $commit = (git ls-remote https://github.com/microsoft/vcpkg.git HEAD | Select-String -Pattern '([a-f0-9]{40})').Matches.Value
            $VCPKG_COMMIT_ID = $commit
        }
        Write-Host "VCPKG_COMMIT_ID=$VCPKG_COMMIT_ID"
        echo "VCPKG_COMMIT_ID=$VCPKG_COMMIT_ID" >> $env:GITHUB_ENV
      env:
        vcpkgRelease: '${{ vars.VCPKG_COMMIT_ID }}'
```

bmarques1995: this for me is ok

walbourn: I can submit this to the main, but you'll need to resolve conflicts or revert these changes.

bmarques1995: I've reverted the changes

bmarques1995: > I can submit this to the main, but you'll need to resolve conflicts or revert these changes.

And because you've explained that the pipelines are expect to not work, is easier to revert them

walbourn: It  was a quirk of the VCPKG usage I didn't resolve before, but this works great. Thanks for the suggestion/implementation.

PR #574, opened 2025-01-31
Update VCPKG GHAs to use head commit id if variable isn't set

walbourn:
In restricted workflows, there's no access to the repository variable that gives the latest stable VCPKG commit id. This adds logic to use the head commit id from the VCPKG registry if the variable isn't set.


walbourn:
Thanks to @bmarques1995 for their help with this.

scm-xbox (approved):

PR #575, opened 2025-02-01
Add option to build as shared lib

bmarques1995:
I accidentally closed the #573, and I'm creating a new PR with the new pipeline check

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,8 @@ project (DirectXTex
 
 option(BUILD_TOOLS "Build tex command-line tools" ON)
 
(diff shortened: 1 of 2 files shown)


walbourn:
Thanks for the contribution. I have some follow-up changes here to add `VERSIONINFO` to the produced DLL, and a few minor whitespace changes. I also want to make sure there's coverage for this scenario so I'll add jobs to some GitHub Actions to make sure this works long-term.

I also should look at DirectXMesh and UVAtlas for the same scenario. It still doesn't make any sense to me to add this to DirectX Tool Kit at this time.

walbourn (approved):

PR #578, opened 2025-02-04
Add VERSIONINFO to DLL version of library

walbourn:
In #575 support for building the library as a DLL using CMake's ``BUILD_SHARED_LIBS`` was added. This PR adds `VERSIONINFO` to the produced DLL, and it is now tested by GitHub Actions.


CMakeLists.txt
```diff
@@ -222,11 +222,19 @@ if(BUILD_DX11 AND WIN32 AND (NOT (XBOX_CONSOLE_TARGET STREQUAL "durango")))
...
+  configure_file(
+      "${CMAKE_CURRENT_SOURCE_DIR}/build/DirectXTex.rc.in"
+      "${CMAKE_CURRENT_BINARY_DIR}/DirectXTex.rc" @ONLY)
+
+  add_library(${PROJECT_NAME} SHARED ${LIBRARY_SOURCES} ${LIBRARY_HEADERS} "${CMAKE_CURRENT_BINARY_DIR}/DirectXTex.rc")
+
+  target_compile_definitions(${PROJECT_NAME} PRIVATE DIRECTX_TEX_EXPORT)
+  target_compile_definitions(${PROJECT_NAME} INTERFACE DIRECTX_TEX_IMPORT)
+else()
+  add_library(${PROJECT_NAME} ${LIBRARY_SOURCES} ${LIBRARY_HEADERS})
```
walbourn, line 237: Note that CMake determines if it uses SHARED or STATIC here based on BUILD_SHARED_LIBS. We handle it  as a special case for Windows to make sure we embed VERSIONINFO. Otherwise the else case handles SHARED on Linux and STATIC on all platforms.

build/versioninfo.ps1
```diff
@@ -7,5 +7,5 @@ param(
 [string]$version
 )
 $versionComma = $version.Replace(".", ",")
-$files = 'Texassemble\texassemble.rc', 'Texconv\Texconv.rc', 'Texdiag\texdiag.rc'
+$files = 'Texassemble\texassemble.rc', 'Texconv\Texconv.rc', 'Texdiag\texdiag.rc', 'build\DirectXTex.rc.in'
```
walbourn, line 10: This is for completeness should I ever decided to add building the DLL version as part of the NuGet pipeline some day.

scm-xbox (approved):

PR #580, opened 2025-02-08
CodeQL recommendations

walbourn:
Improves code quality for the Xbox tiling/detiling auxiliary functions.


scm-xbox (approved):

Auxiliary/DirectXTexXboxDetile.cpp
```diff
@@ -32,7 +32,9 @@ namespace
         const uint8_t* sptr = xbox.GetPointer();
         const uint8_t* endPtr = sptr + layout.SizeBytes;
 
-        for (uint32_t item = 0; item < nimages; ++item)
+        assert((nimages > 0) && (nimages <= UINT32_MAX));
```
walbourn, line 35: I've already checked this condition at the entry-point, but to make sure the assertion holds over time, I've added the assert.

PR #583, opened 2025-02-15
Fixed MinGW with shared libs

walbourn:
* For inline declared function bodies in classes, GNUs parser emits an error with declspec dllimport. The workaround is to add a redundant ``inline``.

* There are also a lot of warnings about ignore attributes on inline functions with dllimport, but for MSVC you need the annotation for correctness.


walbourn:
https://github.com/walbourn/directxtextest/commit/178e8a76d1c38620faabde052ff464651058c585

scm-xbox (approved):

PR #588, opened 2025-02-20
Support loading another DX10 DDS variant with permissive

walbourn:
The DX10 DDS header extension has a field ``arraySize``. For 2D Texture Arrays, this is the number of textures in the array. For cubemaps and cubemapArrays, however, it is supposed to be the number of cubes (i.e. 1 is a single cube with 6 array entries, 2 for two cubemaps with 12 array entries, etc.). This is how D3DX10/D3DX11 write these files. Some writers, however, fail to handle this special case and write the original array count. This PR allows the loading of these files when using the ``--permissive`` option.

Note that the *GetMetadata* functions will return the original values from the header. The *Load* functions will fix up the metadata value returned when this scenario is detected.

### DIFF
--- a/DirectXTex/DirectXTexDDS.cpp
+++ b/DirectXTex/DirectXTexDDS.cpp
@@ -2006,9 +2006,35 @@ HRESULT DirectX::LoadFromDDSMemoryEx(
             return E_FAIL;
     }
 
+    size_t remaining = size - offset;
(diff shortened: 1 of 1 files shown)


walbourn:
https://github.com/walbourn/directxtextest/pull/59

scm-xbox (approved):

PR #593, opened 2025-03-06
CMake update for try-compile with DLLs

walbourn:


scm-xbox (approved):

walbourn:
Turns out this was still needed for legacy durango. Fixed in #597.

PR #594, opened 2025-03-15
IsPlanar option for D3D12 vs D3D11 depth

walbourn:
Depth/Stencil formats are explicitly considered "planar" by Direct3D 12, but were not in Direct3D 11. This PR updates **IsPlanar** with a defaulted optional parameter to handle these two choices, defaulting to Direct3D 11 as before.


walbourn:
https://github.com/walbourn/directxtextest/pull/60

scm-xbox (approved):

PR #595, opened 2025-03-17
Add ComputeTileShape utility

walbourn:
Adds helper to compute the 'shape' of 1D, 2D, and 3D tiles based on the format to get 64KiB sized texture tiles. This uses the 'standard' tiles as defined for Direct3D 11.2 and Direct3D 12, but does not support any Xbox-specific or IHV-specific tiling features.


walbourn:
https://github.com/walbourn/directxtextest/pull/61

scm-xbox (approved):

PR #607, opened 2025-04-13
Update DDSTextureLoader12.cpp

karurung:
The original comment shows there is no DXGI format, so I used `ISBITMASK(0x000000ff, 0x0000ff00, 0x00ff0000, 0)`. This format doesn't have alpha, and I'm sure this is 32 bit. Therefore, in DX12, I think `DXGI_FORMAT_B8G8R8X8_UNORM`, which also helps to go SRGB.

### DIFF
--- a/DDSTextureLoader/DDSTextureLoader12.cpp
+++ b/DDSTextureLoader/DDSTextureLoader12.cpp
@@ -744,7 +744,7 @@ namespace
     //--------------------------------------------------------------------------------------
 #define ISBITMASK( r,g,b,a ) ( ddpf.RBitMask == r && ddpf.GBitMask == g && ddpf.BBitMask == b && ddpf.ABitMask == a )
 
-    DXGI_FORMAT GetDXGIFormat(const DDS_PIXELFORMAT& ddpfm bool& needSwizzle) noexcept
+    DXGI_FORMAT GetDXGIFormat(const DDS_PIXELFORMAT& ddpf, bool& needSwizzle) noexcept
     {
         needSwizzle = false;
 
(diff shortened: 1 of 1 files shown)


DDSTextureLoader/DDSTextureLoader12.cpp
```diff
@@ -768,7 +768,10 @@ namespace
                     return DXGI_FORMAT_B8G8R8X8_UNORM;
                 }
 
-                // No DXGI format maps to ISBITMASK(0x000000ff,0x0000ff00,0x00ff0000,0) aka D3DFMT_X8B8G8R8
+                if (ISBITMASK((0x000000ff,0x0000ff00,0x00ff0000,0))
+                {
+                    return DXGI_FORMAT_B8G8R8X8_UNORM;
```
walbourn, line 773: This bitmask would be `DXGI_FORMAT_R8G8B8X8_UNORM` which doesn't exist.

For this file you need to use **DirectXTex** which has code to do the swizzle and conversions needed to get it into a DXGI format.

walbourn: Note that ``DXGI_FORMAT_B8G8R8X8_UNORM`` mask is already in the code:

```
// ddpf.RGBBitCount == 32
ISBITMASK(0x00ff0000, 0x0000ff00, 0x000000ff, 0)
```

karurung: Hi, I tried to add conversion and swizzling to change pixel format, which is RGBA to BGRA. Is there anything I need to change?

walbourn:
The general philosophy is that **DDSTextureLoader** will load all content that directly maps to a DXGI format. For handling all other formats that require pixel-format conversion, filtering, swizzling, or any other alterations to the data.

What makes this scenario make sense to add special-case handling for the otherwise 'light-weight' loader?

karurung:
I was a bit confused, I thought the swizzling logic should go in `DDSTextureLoader`, but now I understand that you're saying it should be in `DirectXTex` instead thanks for clarifying!

PR #631, opened 2025-09-17
Upgrade to CMake 3.21

walbourn:
CMake 3.21 comes with VS 2022.

> VS 2019 users will need to install a standalone version of a CMake newer than the 3.20 that comes with it.


walbourn:
Test suite update https://github.com/walbourn/directxtextest/commit/26ff6d74b3ce6ffc060788bd7b25ea9662fe8c1b

nadaOuf (approved):

PR #645, opened 2025-10-31
Add support for LoadFromEXRMemory as other formats

vkaytsanov:
Tested from memory and from file the following exr: https://polyhaven.com/a/qwantani_dusk_2_puresky

Changes:
Rename *Stream -> *FileStream
Add InputStream wrapper to read from memory
Cut-paste the whole LoadFromEXRFile try-catch into LoadFromEXRCommon
Remove the equal sign from COMBINED_OPENEXR_VERSION  as I am exactly on this version and I don't have this virtual method, todo: implement it in the future

### DIFF
--- a/Auxiliary/DirectXTexEXR.cpp
+++ b/Auxiliary/DirectXTexEXR.cpp
@@ -255,7 +255,7 @@ namespace
 #endif // _WIN32
 
     //-------------------------------------------------------------------------------------
-    // Load 
+    // Load
     //-------------------------------------------------------------------------------------
     template <typename StreamType>
     HRESULT LoadFromEXRCommon(StreamType stream, _Out_opt_ TexMetadata* metadata, ScratchImage& image)
@@ -414,6 +414,7 @@ HRESULT DirectX::GetMetadataFromEXRFile(const wchar_t* szFile, TexMetadata& meta
             }
         }
 
+        metadata = {};
         metadata.width = static_cast<size_t>(width);
         metadata.height = static_cast<size_t>(height);
         metadata.depth = metadata.mipLevels = 1;


walbourn:
Can you rebase this?

vkaytsanov:
@walbourn rebased and rearranged the code a little bit to give better diff. Please note, that I haven't tested non-windows desktop systems for the template argument deduction

Auxiliary/DirectXTexEXR.cpp
```diff
@@ -320,60 +360,25 @@ HRESULT DirectX::GetMetadataFromEXRFile(const wchar_t* szFile, TexMetadata& meta
...
 
-
 //-------------------------------------------------------------------------------------
-// Load a EXR file from disk
+// Load 
 //-------------------------------------------------------------------------------------
 _Use_decl_annotations_
-HRESULT DirectX::LoadFromEXRFile(const wchar_t* szFile, TexMetadata* metadata, ScratchImage& image)
+template <typename StreamType>
+HRESULT LoadFromEXRCommon(StreamType stream, TexMetadata* metadata, ScratchImage& image)
```
walbourn: This template is not public so it should be placed in an anonymous namespace.

`_Use_decl_annotations_` is not valid here and instead it needs to be SAL annotated directly. In this case, it's sufficient to make it:

```
HRESULT LoadFromEXRCommon(StreamType stream, _Out_opt_ TexMetadata* metadata, ScratchImage& image)
```

vkaytsanov: The diff was really bad when it was in the anonymous namespace. Can move it after approval with the removal of `_Use_decl_annotations_`

solbjorn:
Any plans to add support for loading mipmaps from tiled exrs?

PR #669, opened 2026-02-11
VS 2026 projects

walbourn:
Adds VS 2026 slnx and project files, as well as GHA coverage for MSVC compiler 19.50 and clang 20.1.8.

Note I call it `2026` and not `2026_Win10` because VS 2026 doesn't support targeting Windows 8.1 or earlier.

> Includes some code review feedback for some scripts as well.


walbourn:
Rather than add a set of 2026 slnx/projects for the Test Suite, I just removed all the pipelines and use of the `Tests_` in favor of CTest via CMake.

https://github.com/walbourn/directxtextest/pull/68

scm-xbox (approved):

PR #671, opened 2026-02-12
Added bit bias when storing R8_UNORM to be consistent with other 8bpp

walbourn:
Adds a bias in StoreScanline when converting the float values to 8bpp channel specifically for `DXGI_FORMAT_R8_UNORM` and `DXGI_FORMAT_A8_UNORM`.  This makes it consistent with other 8bpp formats.


walbourn:
Test updates

https://github.com/walbourn/directxtextest/pull/69

PR #674, opened 2026-02-26
Add `LoadFromDDSStream{,Ex}()`

solbjorn:
Many games support own archive formats and implement own sort of a virtual file system. In that case, `LoadFromDDSFile()` doesn't fit, while `LoadFromDDSMemory()` requires 2x filesize: one buffer to read the file and another one to load the texture, since `ScratchImage` is always owning.

Add a simple `InputStream` class and `LoadDDSFromStream()` to avoid this memory overhead and allow games and apps to use their own functionality to operate with files or memory.
More formats could support this in future, as well as writing to an `OutputStream` if/when needed.

### DIFF
--- a/Auxiliary/DirectXTexEXR.cpp
+++ b/Auxiliary/DirectXTexEXR.cpp
@@ -82,10 +82,10 @@ namespace
         HRESULT result;
     };
 
-    class InputStream : public Imf::IStream
+    class EXRInputStream : public Imf::IStream
     {
     public:
-        InputStream(HANDLE hFile, const char fileName[]) :
+        EXRInputStream(HANDLE hFile, const char fileName[]) :
             IStream(fileName), m_hFile(hFile)
         {
             const LARGE_INTEGER dist = {};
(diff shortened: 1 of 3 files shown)


solbjorn:

walbourn:
For scenarios where you are just wanting to load content, DirectXTex is going to be overkill unless your dataset includes lots of legacy formats (24bpp, etc.) you have to fix-up at load-time.

For most titles/engines, DirectXTex is an excellent library to use for a 'content cooking/packing' step that produces a 'WAD-style' file-system-in-a-file' data file that your application can efficiently load with ASYNC I/O. Loading them as individual DDS files generally gets dominated by the overhead of Open/Close and AV scans.

Is there a particular design requirement here that your engine consume 'loose' files?

walbourn:
And to verify for your scenario, you DO NOT expect the loader at runtime to do legacy conversions/expansions (24bpp, 3:2:2, etc.)?

solbjorn:
We only use `DirectX::CaptureTexture()` from DirectXTex to get our main render target and later we process and save the image manually, as we have to handle different types of screenshots (regular screenshots are in PNG or JPG and for this, ScreenGrab would be enough; we also have thumbnails for game saves, which are saved in DDS, but resized and compressed in BC7). So it's not a big deal which API to use for that (and screenshots is not something perf-critical), although we usually write files using our engine API for consistency.

PR #676, opened 2026-02-27
HDR file input validation fixes

walbourn:
* Fix handling for specific truncated header that otherwise results in overreading a input buffer.

* Fix detection of supported vs. unsupported orientation (only `-Y height +X width` is supported).


nadaOuf (approved):

walbourn:
Resolves MSRC 107429 which is considered a Low risk Information Disclosure bug.

PR #684, opened 2026-03-20
Add support for powerpc64[le] systems

madscientist159:
This requires POWER ISA 3.0 or higher
Older ISA variants lack current deployment and required features


walbourn:
Would you be willing to make the same PR to DirectXMesh and see if it builds for you?

https://github.com/microsoft/DirectXMesh

